PR #82, opened 2016-05-25
Expand Travis CI tests

pllim:
Added `egg_info`, PEP8, and coverage tests. Check for any unclosed files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ env:
         - CONDA_DEPENDENCIES='Cython jinja2 scipy'
     matrix:
         # Make sure that egg_info works without dependencies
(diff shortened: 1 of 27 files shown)


embray:
This should probably use the same external module loading mechanism that astropy itself uses for `six`.  Otherwise anyone who packages synphot for Debian is going to run into the same issue astropy had, that it should be able to use the system version of the `six` module.

(Maybe this should just be part of the astropy package-template)

pllim:
@embray , not very sure what you mean. I got this copy of `six` from Ginga and it is on Debian (or almost there).

pllim:
p.s. Why is it not in `astropy_helpers`? Could have saved me some head scratching.

PR #84, opened 2016-06-02
Use Astropy mag units

### LINKED ISSUE #32: Use astropy magnitude system when available
In ``astropy.units.astrophys`` documentation:

> The mag unit is provided for compatibility with the FITS unit string standard. However, it is not very useful as-is since it is “orphaned” and can not be converted to any other unit. A future astropy magnitudes library is planned to address this shortcoming.

So, when this library is available, pysynphot should be updated to use that library instead of local definition of magnitudes.

pllim:
- [x] Replaced STMAG and ABMAG with official units from Astropy.

This fixes #32. It is unlikely that VEGAMAG and OBMAG will ever be in Astropy, as they depend on standard star/telescope area, so just leave those as-is.

Note: When astropy/astropy#5017 is merged, `spectral_density_mag` can be removed.


pllim:
@eteq, to answer your question from a while ago... yes, the Astropy logarithmic units seem to be working well. This PR demonstrates how I was able to incorporate them into `pysynphot` refactoring. 

Can't figure out an easy way to add my own units (i.e., `obmag` and `vegamag`) that needs extra info like telescope area or Vega spectrum file, but that has been discussed before in Astropy issues, so we don't have to do that again here.

PR #100, opened 2016-08-30
Change default data encoding to binary

pllim:
Removed the need to specifically provide encoding to read commonly supported data.


pllim:
:tada: 100! :tada:

PR #113, opened 2016-11-28
Improve plot method with Quantity

pllim:
Improve `plot()` method with Quantity. The code was written when Quantity was not supported by `matplotlib` but things have changed.


pllim:
Not waiting for Travis since plotting is not tested there. Manually tested already.

Issue #71, opened 2014-08-18
Enable auto doc via Read the Docs


pllim:
The link is registered here:
https://readthedocs.org/projects/synphot/

Depends on #63 .

pllim:
Done!

PR #114, opened 2016-11-30
Write complete user's guide

pllim:
Port relevant portions of `pysynphot` user's guide over, in addition to new materials.

This will lead to #71.


pllim:
Merging now so I can test RTD hook.

Issue #125, opened 2017-07-21
Fix failed tests due to Astropy constants update


pllim:
See astropy/astropy#6383. The easy way out is simply to update the expected values to be the new ones. However, this might break the commissioning tests because Astrolib PySynphot has no concept of different versions of constants (it simply uses the older version).

PR #127, opened 2017-08-03
Fix negative leak

### LINKED ISSUE #126: Count rate with range computes sum that exceeds total count rate
Similar problem as spacetelescope/pysynphot#43 . Need to change https://github.com/spacetelescope/synphot_refactor/blob/master/synphot/observation.py#L561 to use `math.fsum()` and add a test to make sure it does not happen again.

Note: The example uses `stsynphot` specific code, but the bug is here.

```python
import stsynphot as STS
from synphot import Observation
sp = STS.parse_spec('rn(icat(k93models,44500,0.0,5.0),box(2000.000000,1.),1e-10,flam)')
bp = STS.band('stis,fuvmama,mirror,F25NDQ2,MJD#58300.0822774')
obs = Observation(sp, bp, binset=bp.binset)
c_all = obs.countrate(STS.conf.area)  # <Quantity 123649.12208258784 ct / s>
c_sub = obs.countrate(STS.conf.area, waverange=[1160.22, 12000.0])  # <Quantity 123649.12208258786 ct / s>
assert c_sub <= c_all  # AssertionError!!!
```

Fun fact: It is not easy to find a test case that raises this error using analytical models!

pllim:
Fix #126


pllim:
Failures are due to #125 and unrelated to this PR.

PR #151, opened 2018-06-01
Fix units

### LINKED ISSUE #149: Unit from integrate() is wrong
For example, instead of FLAM (`erg/s/cm^2/A`), it should be just `erg/s/cm^2` because `A` is cancelled out by integration over wavelength. I need to:

1. Fix `integrate()` output unit.
2. Check to make sure other methods have correct units.

This was an oversight because PySynphot results did not attach real units to their values, so I had to revisit the original papers and probably misinterpreted this one. (This is an issue over at PySynphot too but it is a documentation issue rather than actual bug.)

This arose from a help call (INC0060889).

I apologize for any inconvenience caused. :cry:

pllim:
Fix #149

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -193,13 +193,13 @@ Sample the spectrum at 0.3 micron::
 
     >>> from astropy import units as u
     >>> sp(0.3 * u.micron)
-    <Quantity 0.001268063356632316 PHOTLAM>
+    <Quantity 0.0012685 PHOTLAM>
 
 Models that built the spectrum::
 
     >>> print(sp)
     SourceSpectrum at z=0.0
-    Model: CompoundModel5
+    Model: CompoundModel2
     Inputs: ('x',)
     Outputs: ('y',)
     Model set size: 1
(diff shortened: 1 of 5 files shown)


docs/index.rst
```diff
@@ -233,7 +233,7 @@ integrate it::
 
     >>> sp_rn = sp.normalize(1 * u.Jy, band=bp)
     >>> sp_rn.integrate()
-    <Quantity 12900.947986966807 PHOTLAM>
+    <Quantity 12901.13466265 ph / (cm2 s)>
```
pllim: @eteq does this make sense to you? I just need a sanity check. Thanks!

synphot/tests/test_spectrum.py
```diff
@@ -660,16 +671,6 @@ def test_renorm_vegamag(self, sp_type, ans_countrate):
                              vegaspec=_vspec)
         self._compare_countrate(rn_sp, ans_countrate)
 
-    def test_renorm_noband_count(self):
```
pllim: I really don't remember why I added this test in the first place... :grimacing:

pllim:
Note to self: Backport and bugfix release after merge. :scream:

PR #166, opened 2018-10-19
Make effstim calculation consistent with PySynphot

### LINKED ISSUE #159: Conversion between wavelength and frequency giving discrepant results
Testing some synthetic photometry on Vega, I'm getting different results between `pysynphot` and `synphot`. I've verified that both `pysynphot` and `synphot` are using the same Vega spectrum and the bandpass ("acs,wfc1,f606w") is using the same files and plotting the filter throughput curves they look identical.

```
import pysynphot as psyn
import synphot as syn
import stsynphot as stsyn
import matplotlib.pyplot as plt
import numpy as np
import os
from astropy import units as u

# Set up bandpass
bandpass = 'acs,wfc1,f606w'
pband = psyn.ObsBandpass(bandpass)
sband = stsyn.band(bandpass)

# Set up vega spectra
pvega = psyn.FileSpectrum(os.path.join(os.environ['PYSYN_CDBS'], 'calspec', 'alpha_lyr_stis_008.fits'))
svega = stsyn.Vega

# Make observations
pvega_obs = psyn.Observation(pvega, pband)
svega_obs = syn.Observation(svega, sband, binset=sband.binset)

# Get fluxes
pflux = pvega_obs.effstim('flam')
sflux = svega_obs.effstim(syn.units.FLAM)

pflux_jy = pvega_obs.effstim('jy')
sflux_jy = svega_obs.effstim(u.Jy)
```

I've done this several different ways now, and find the following differences compared to the `pysynphot` result:

| Type            | PySynphot | Synphot | % Difference |
|-------------|--------------|---------------|-------------|
| Count rate | 36395078438.946 | 36395792271.484 | 0.00196 |
| FLAM | 2.848e-9 | 2.848e-9 | 1.4531e-14 |
| Jy | 3331.797 | 3251.440 | 2.4118 |
| FNU | 3.332e-20 | 3.251e-20 | 2.4118 |
| STmag | 0.264 | 0.264 | 2.3166e-12 |
| ABmag | 0.093 | 0.120 | 28.4093 |

Changing the obsmode (trying ACS/WFC1 F435W, F606W, F814W, and WFC3/UVIS1 F218W, F814W) gives varying results but the frequency-based values are seemingly always off by ~1-2.5% (except ABmag, which has jumped between 2–28%).

Unless I'm doing something wrong with the my commands for calculating the flux in each unit (do I need to do an extra step when using frequency-based units?), then it seems like there's something wrong with the conversion between wavelength and frequency in one of the two packages, but it's not clear which one. @pllim suggested opening this issue on the `synphot` repo so that it can be investigated (and also suggested @bhilbert4 might be interested).

pllim:
Fix #159

I can't say I understand why PHOTNU needs a special logic block. I tried converting from PHOTLAM and FLAM using bandpass pivot (like what is done with FNU etc) but it didn't work out. At least all the supported flux units give values consistent with PySynphot now (in my limited testing). 

Results *before* the fix: https://gist.github.com/pllim/2d053b327d384c587b0297950644e7d1

Results *after* the fix: https://gist.github.com/pllim/c378494302ad4df095fe506df8ae8ade

**TODO**

- [x] Wait for resolution of spacetelescope/pysynphot#103 and update this PR, if necessary
- [x] Add/fix tests
- [x] Add change log
- [x] Update examples in doc, if necessary
- [x] Wait for Tyler's feedback

cc @deustua @tddesjardins @bhilbert4 (Tyler, do you want to take this for a spin?)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@
 
 - Config to use HTTP instead of FTP. [#171]
 - Fixed scalar unit conversion for VEGAMAG. [#174]
+- Bug fix for ``effstim`` calculations in some flux units. [#159, #166]
 
 0.1.2 (2018-07-19)
 ==================
--- a/docs/synphot/formulae.rst
+++ b/docs/synphot/formulae.rst
@@ -283,14 +283,16 @@ The default binning behavior is to be consistent with ASTROLIB PYSYNPHOT.
 
 Example::
 
-    >>> obs.effstim()  # Not binned
-    <Quantity 0.00054170149051543 PHOTLAM>
+    >>> obs.effstim()
+    <Quantity 0.00053744 PHOTLAM>
     >>> obs.effstim('flam')
-    <Quantity 1.992237048596971e-15 FLAM>
-    >>> obs.effstim('count', area=area, binned=True)  # Binned
-    <Quantity 6624.720529866574 ct / s>
-    >>> obs.countrate(area=area)
-    <Quantity 6624.720529866574 ct / s>
+    <Quantity 1.99333435e-15 FLAM>
+    >>> obs.effstim('count', area=area)  # Not binned
+    <Quantity 6628.36886854 ct / s>
+    >>> obs.countrate(area=area, binned=False)
+    <Quantity 6628.36886854 ct / s>
+    >>> obs.countrate(area=area)  # Binned
+    <Quantity 6628.36888121 ct / s>
 
 
 .. _synphot-formula-effwave:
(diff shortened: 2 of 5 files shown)


pllim:
I, uh, actually am suspecting that PHOTNU calculation in PYSYNPHOT is *wrong* (hence this PR is also on hold until this suspicion can be lifted). Halp...

pllim:
Oooo `coveralls` works again!

tddesjardins:
Tested this with all of the ACS/WFC filters and using different source spectra and two MJDs using the FLAM, FNU, ABmag, and STmag units. Results agree to <1E-6 mags between synphot and pysynphot, and with my simple mathematical calculations to about 1E-4 magnitudes.

PR #171, opened 2018-12-12
Config to use HTTP instead of FTP

### LINKED ISSUE #168: TST: Fix compatibility issues with pytest 4.0.1
`pytest` 4.0.1 emits some new warnings, which is causing tests to fail in both Travis and Appveyor. Have to fix them, or if impossible, ignore those warnings.

pllim:
Fix #168 in the process. Goal: Get badges to be green again.


pllim:
I am so tempted to talk to myself here because no one else is going to review this.

PR #184, opened 2019-03-25
EXP: See why astropy dev fails

pllim:
Travis CI did not kick off over at #183, so trying my luck in a new PR.


pllim:
I switched to Azure, so I have no motivation to pursue this.

PR #185, opened 2019-03-31
Remove Python 2 support, misc clean-ups

### LINKED ISSUE #131: Drop Python 2 support
To be consistent with Astropy dev and v3.0 release later this year. Since this package is not very widely used yet, I don't see any problem with this.

pllim:
Fix #131 , plus other clean-ups.


pllim:
Oh yeahhhh

PR #190, opened 2019-05-13
[WIP] Azure CI 

### LINKED ISSUE #181: TST: Move to Azure
With the move to Azure, both Travis and Appveyor can be retired, reducing maintenance effort in the long run.

Also need to add a doc build check. How is it not already in CI?!

pllim:
Xref #181

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,7 +7,7 @@ trigger:
 - master
 
 jobs:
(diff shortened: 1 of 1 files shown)


pllim:
@jhunkeler , Azure is not running. Is there something you need to enable as org admin for `spacetelescope`? 🙏

Issue #188, opened 2019-04-15
download_data does not download anything - skips everything


pmaxted:
Trying to install synphot following instructions at https://synphot.readthedocs.io/en/latest/#installation-and-setup
```
% pip install synphot
Successfully installed synphot-0.1.3

% cd ~/.astropy/config/

% vi synphot.cfg
> [config]
>
> ## Default integrator
> default_integrator = 'trapezoid'
>
> ## Vega
> vega_file = /Users/pflm/cdbs/calspec/alpha_lyr_stis_008.fits
etc...

% mkdir  /Users/pflm/cdbs/
% ipython
In [1]: from synphot.utils import download_data
In [2]: file_list = download_data('/Users/pflm/cdbs')
/Users/pflm/cdbs/calspec/alpha_lyr_stis_008.fits is not from http://ssb.stsci.edu/cdbs/, skipping download
/Users/pflm/cdbs/extinction/lmc_30dorshell_001.fits is not from http://ssb.stsci.edu/cdbs/, skipping download
```
etc..

pllim:
Temporarily rename you `synphot.cfg` to, say, `synphot.cfg.bak` first. Then try `download_data` again. The workflow I had imagined was people would download it first, and then only create the `synphot.cfg` file to point to downloaded data.

I guess the doc needs to be re-arranged a little to avoid confusion.

eteq:
Note that I added some text as a part of #208 that explains what's happening here.  Maybe that's sufficient to declare this closed when/if it's merged?

The alternative would be a more radical change to `download_data` that actually separates the "source" and "destination" into two separate config items.  Or a slightly softer version would be to instead assume the default config item value is the only "true" url and always use that for download. (which would just be a one-line change I think).

pllim:
Yes, perhaps a logic like "if this thing from synphot.cfg does not exist, then download from the default" would work, similar to the "softer" version that @eteq suggested above. I am :-1: on having more config values, as they can get confusing. This should be done after #208 is settled though, to avoid conflict.

pllim:
@pmaxted , a fix for this was done in #211 and will be out in a future release. We apologize for any inconvenience caused.

PR #191, opened 2019-05-24
Reworded the bit about downloading data for pre-defined quantities

tcjansen:
In response to #188 and my own confusion, I've reworded the section on downloading data for using the pre-defined standard star, extinction laws, and bandpasses in the docs.


tcjansen:
Ah I see! Okay, I've rearranged the wording a bit more.

PR #192, opened 2019-05-28
Tutorials [work in progress]

tcjansen:
Here is where we can comment on and make changes to some synphot tutorials!


tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# <center>Computing the count rate of observing HAT-P-11 with the APO 3.5m telescope using ``synphot``</center>"
```
pllim: I don't see this title rendered when I click "View File". Hmm.

Also, if we really go the notebook route, there needs to be a brief description of what the star is, or maybe simply linking it to https://en.wikipedia.org/wiki/HAT-P-11 .

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+      "text/plain": [
+       "<Figure size 432x288 with 1 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "T_eff = 4780 * u.K  # temperature of hatp11 [K]\n",
```
pllim: If you obtained this value from some literature or website, please cite it. Same comment applies to other numbers used below.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "\n",
+    "# Use synphot.models and synphot.spectrum module to generate a blackbody spectrum\n",
+    "photlam_hatp11 = SourceSpectrum(BlackBody1D, temperature=T_eff) # in photon s-1 cm^-2 AA^-1\n",
```
pllim: The comment `in photon s-1 cm^-2 AA^-1` is misleading here. At this point, `photlam_hatp11` is an instance of `SourceSpectrum`, which _contains_ the `BlackBody1D` model describing the star. As far as user is concerned, flux unit does not come into play until you sample the spectrum.

In addition, usually when creating a theoretical source spectrum, one normalizes it to some reference flux in a reference filter. For example, I can create a blackbody with this temperature, but then I normalize the blackbody to have the property of, say, 28 ABmag when observed in, say, *V*-band (this depends on existing observational data of your object of interest). A source spectrum without normalization like this sometimes give crazy results.

See also: [synphot.spectrum.BaseSourceSpectrum.normalize](https://synphot.readthedocs.io/en/latest/api/synphot.spectrum.BaseSourceSpectrum.html#synphot.spectrum.BaseSourceSpectrum.normalize)

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "\n",
+    "# Compute the flux at Earth from the luminosity \n",
+    "R = 0.683 * u.R_sun  # radius in solar units\n",
+    "d = (123 * u.lyr).to(u.R_sun)  # distance to HATP11 in solar radii\n",
+    "photlam_hatp11 = photlam_hatp11 * (R / d) ** 2\n",
+    "\n",
+    "photlam_hatp11.plot(left=3000, right=10000)"
```
pllim: When plotting, user usually connects more with flux unit like FLAM or Jy. You can use `flux_unit` to set the displayed flux unit, as documented in [synphot.spectrum.SourceSpectrum.plot](https://synphot.readthedocs.io/en/latest/api/synphot.spectrum.SourceSpectrum.html#synphot.spectrum.SourceSpectrum.plot) .

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from synphot import SpectralElement\n",
+    "\n",
+    "agile = SpectralElement.from_file('agile.dat', wave_unit=\"nm\")\n",
```
pllim: I think @eteq has some strong opinion about the API here. Erik, feel free to open a new issue to discuss it.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from synphot import Observation\n",
+    "import numpy as np\n",
+    "\n",
+    "obs = Observation(photlam_hatp11, agile)\n",
```
pllim: Here, assuming that the throughput file is binned correctly already, you can pass in `binset=agile.waveset`. Alternately, you may also inquire with APO for the values, and then use `numpy` to construct them as array before passing it into `binset`.

For example, here is a `binset` file for HST/COS using one of its grism filter: [cos_fuv_g130m_1291.txt](https://github.com/spacetelescope/stsynphot_refactor/blob/master/stsynphot/data/wavecats/cos_fuv_g130m_1291.txt)

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "11852206.200075638 ct / s\n"
+     ]
+    }
+   ],
+   "source": [
+    "r = 3.5 / 2 * u.m  # radius of 3.5m ARC telescope at APO\n",
+    "area = np.pi * r.to(u.cm) ** 2\n",
```
pllim: Technically, `r.to(...)` is unnecessary. As you are passing in a Quantity, there is unit attached already and `synphot` should know how to handle it under the hood. And if you find out that it cannot, then it is a bug.

```python
area = np.pi * r ** 2
```

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "\n",
+    "cr = obs.countrate(area=area)\n",
+    "print(cr)"
```
pllim: `11852206.200075638 ct / s` seems like a lot to me, but then again I am not familiar with this object. @bmorris3 should check if this number makes sense.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+   "source": [
+    "T_eff = 4780 * u.K  # temperature of hatp11 [K]\n",
+    "\n",
+    "# Use synphot.models and synphot.spectrum module to generate a blackbody spectrum\n",
+    "photlam_hatp11 = SourceSpectrum(BlackBody1D, temperature=T_eff) # in photon s-1 cm^-2 AA^-1\n",
+    "\n",
+    "# Compute the flux at Earth from the luminosity \n",
+    "R = 0.683 * u.R_sun  # radius in solar units\n",
+    "d = (123 * u.lyr).to(u.R_sun)  # distance to HATP11 in solar radii\n",
+    "photlam_hatp11 = photlam_hatp11 * (R / d) ** 2\n",
```
pllim: I am not sure how much this replaces the normalization I mentioned above. I guess a sanity check from @bmorris3 will tell us.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "\n",
+    "cr = obs.countrate(area=area)\n",
+    "print(cr)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "How many photons can be expected in a 10 second exposure?"
```
pllim: A more interesting question for an observer (at least what I would want to know back in my observing days) is, "How much exposure time is needed if my target SNR is 10?" (You can replace 10 with what makes sense for your science here.) I think at this point, you have enough info to calculate that?

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,372 @@
...
...EQPCx7+fghXD5+CKcNTdeNHdLm2j3BmFkKsAG4HCgFFgPXu/vamDpfAya5+1fM\nbDZwrbt/zszGA48BU4Bs4FVgnLvXmdlFQDXwUKMEcwUw391rzewuAHf/XksxdpQEs+9gDb94dSMP\nL9xK754p3HbZOP72vFO69FP28nHuzvu7qnk5TDbLPgi+W43I7M1VE4fy6YnDmJjTX8lG2kQyPGg5\nBShx901hQHOBGcDamDozgB+Fy08C91jwCZgBzHX3I8BmMysJ21vg7m+aWV7jN3P3l2NeLgSua9O9\naQe1dfU8uugDfv7KBg4cOsr1U0bwncvHMVDPSkgjZsaYwemMGZzO1y4eQ0XVYV57r4IXV+/g/rc2\n819vbCInI41pE4dy1cRh5A/PULKRyEWZYHKAbTGvS4FzmqsTnnnsBwaG5QsbbZvTive+GXi8tQEn\nk5Wl+/jukytZt6OK80cP5AdXj+f0Yf3aOyzpIAan9+L6KSO4fsoI9h2s4ZW1O3lhVTn/884WfvfW\nZrL792LqGcOYNnEoBSMGaPgaiUSnGyrGzL4P1AKPNLP+VuBWgBEjRpzEyBJz+Ggdd7+ygd+9tYlB\n6an89gsFXDlhqL5tynHL6N2TzxYO57OFw9l/6CivvRckm/+3cCsPvL2ZIf1SueqMYVx1xlAK8zJJ\nUbKRNhJlgikDhse8zg3LmqpTambdgf4Enf2JbPtXzOwmghsALvVmOpfc/T7gPgj6YBLZkZNl0eZK\nvvfHlWze/SGzzx7OHdNOp3+abj+VttM/rQezCnKZVZBL1eGjzF9XwfMry3l00Qf8zztbGJSeytQJ\nQ7lq4lDOGTlQyUZOSJQJZjEw1sxGEiSH2cDnG9UpAm4EFhD0mcx3dzezIuBRM/s5QSf/WGBRS28W\n3rH2XeCT7n6wTfckYtVHavnpn9bx0IKtDM9M45EvncMFY7LibyhyAtJ79WDGmTnMODOH6iO1vL6u\nghdXl/OHJdt4eOFWBqWn8umJw7jmzGz12chxifo25WnALwhuU37A3f/NzO4Eit29yMx6AQ8D+UAl\nMDvmpoDvE/Sl1AK3ufuLYfljwMVAFrAT+KG73x/eCJBKcAYEsNDdv9JSfMlwF9mSrZV887HlbN9/\niJvOz+OfrjyV3j073ZVL6UAO1tTy+rpdFK0o4/X1u6ip
```
pllim: This cell is pretty much a repetition of the earlier source spectrum setup for a different star. I am on the fence on whether to recommend turning this cell into a function in this notebook (not in the package library), and then call the function twice in this notebook, or not.

pllim: Might also want to vary the line styles, to accommodate people with color blindness and such.

pllim: Alternate way of looping can be:

```python
for band_i, crs_i in zip(bands, crs):
    ...
```

Also, @bmorris3 should do a sanity check on these numbers, if possible.

tcjansen:
Thanks for the comments @pllim! I will get to them soon

tutorials/observing_examples.ipynb
tcjansen: I don't see it either, but it loads just fine from the terminal. I think it's a github thing..

pllim: My hunch is that the `<center>` tags are confusing GitHub, but that is not important right now.

tutorials/observing_examples.ipynb
tcjansen: I did see that you normalized your spectra in the examples, but it's unclear to me why this is necessary. Is it just for comparison's sake?

pllim: Normalization is needed when the default model flux values are arbitrary. For example, blackbody flux at a given temperature is the theoretical blackbody from Planck's law, which does not often correspond to real life observations. Say, for the Sun, you would want to normalized that blackbody at solar temperature to the observed solar magnitude at a given filter. Does this make sense?

pllim: Maybe @eteq or @bmorris3 have the official academic explanation.

tutorials/observing_examples.ipynb
tcjansen: Good to know! It does handle it.

tutorials/observing_examples.ipynb
tcjansen: Agreed

tutorials/observing_examples.ipynb
tcjansen: I agree - maybe this is a kind of function we can implement as a part of my project?

tutorials/observing_examples.ipynb
bmorris3: I could be really wrong here, but my current understanding is that the generic blackbody flux in the `BlackBody1D` model should be the flux _at the surface_ of the source. So you need to normalize it by the square of the radius over the distance to the target in order to get the flux at Earth. Does that sound reasonable @pllim?

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,362 @@
...
+   "source": [
+    "T_eff = 4780  # temperature of hatp11[K], Bakos et al. 2010\n",
+    "\n",
+    "# Use synphot.models and synphot.spectrum module to generate a blackbody spectrum\n",
+    "photlam_hatp11 = SourceSpectrum(BlackBody1D, temperature=T_eff)\n",
+    "\n",
+    "# Compute the flux at Earth from the luminosity \n",
+    "R = 0.683 * u.R_sun  # radius in solar units, Deming et al. 2011\n",
+    "d = (123 * u.lyr).to(u.R_sun)  # distance to HATP11 in solar radii, from Gaia parallax\n",
+    "photlam_hatp11 = photlam_hatp11 * (R / d) ** 2  # in photon s-1 cm^-2 AA^-1\n",
```
bmorris3: To avoid the explicit unit conversion that you did for `d`, simple define `d = 123 * u.lyr`. Then in the next line, call `photlam_hatp11 * float(R / d) ** 2`. The `float` call will resolve the unit differences and return the float conversion factor.

tutorials/observing_examples.ipynb
tcjansen: Hmm, I could do, though I wouldn't want a user thinking it's necessary to include, since in practice they could just normalize instead of scaling by r/d^2.  Maybe in this second example I could show a normalization instead of the r/d^2 scaling?

tutorials/observing_examples.ipynb
tcjansen: Right, I thought scaling by r/d^2 would cover the normalization. As I mentioned somewhere below it might be good to use the normalization function for the second example to show both ways?

pllim: Yeah, maybe what you did is already correct. I am not sure coz the count rate looks suspiciously high to me. Or maybe this is just a super bright star, even for APO. 🤷‍♀

pllim: Anyways, sorry for the confusion. I think first of all, we need to check if count rate is correct. Then if not, we can revisit how the normalization is done.

pllim:
p.s. For future reference, you can insert `[skip ci]` in your commit message to skip running the CI. There is no need to run it since it is not set up to test notebooks. Thank you for your patience!

tutorials/observing_examples.ipynb
bmorris3: This example needs to be tweaked to use ARCTIC instead of Agile for comparison with my observations. But it turns out this count rate is correct to within a factor of 2! (You get TONS of counts when observing the brightest exoplanet host stars with a 3.5 m telescope).

tutorials/observing_examples.ipynb
bmorris3: For Poisson noise-limited photometry, the S/N is just `1/sqrt(counts)`, no? If so, I'm not sure we need to write a special method to do that math.

tutorials/observing_examples.ipynb
pllim: Is factor of 2 good or bad? 😸  What is the empirical value from real observation? I am curious!

bmorris3:
The "answer key" is available [here](https://nbviewer.jupyter.org/gist/bmorris3/6165edf1cd671aaf97970df62dd37de0), where I do the most basic photometry possible to get answers. They are: 

> HAT-P-11 with APO/ARCTIC+holographic diffuser: 34549176.0 counts in 10.0 s in SDSS r

and

> TRAPPIST-1 with APO/ARCTIC: 203230.0 counts in 10.0 s in SDSS z

Note that this is slightly different from what's being done in the example presently.

docs/index.rst
```diff
@@ -89,6 +89,11 @@ look for these files.
     **stsynphot**. If you plan to use the latter anyway, please also read the
     instructions in its documentation.
 
+To install local data files via HTTP into ``/my/local/dir/cdbs``::
```
pllim: This does not seem relevant to the notebook?

tutorials/observing_examples.ipynb
pllim: Hmm... `N = sqrt(counts)`, so `counts / sqrt(counts) = sqrt(counts)` is the SNR, and not `1/sqrt(counts)`? That is a naive assumption though. With telescopes, you also need to factor in the sky and readnoise. Up to you on how accurate you want it to be.

Even if a special function is needed, we can prototype it in the notebook first, as mentioned during the tag-up.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,374 @@
...
...7jwH2hG2LiEg7ifIMZjJQ6u4b3L0GmA1Mb1RnOvBwuPw0cIWZWVg+292PuPtGoDRsr8k2w20uD9sgbPP6CPdNRETiiDLB5AJbY16XhWVN1nH3WmAfkNXCts2VZwF7wzaaey8RETmFokww1kSZJ1inrco/GZTZbWZWYmYlO3fubKqKiIi0gSgTTBkwLOZ1HrCtuTpm1h3oD1S1sG1z5buAjLCN5t4LAHd/wN2L3L1o4MCBJ7BbIiKSiCgTzAJgTHh3V0+CTvviRnWKgZvD5RuAue7uYfms8C6zEcAYYH5zbYbbvBW2Qdjm8xHum4iIxNE9fpUT4+61ZvZt4FUgBXjI3Vea2d1AibsXAw8Cj5pZKcGZy6xw25Vm9hSwCqgFvuXudQBNtRm+5R3AbDP7F2Bx2LaIiLQTC778d01FRUVeUlLS3mGIiHQoZrbQ3Yvi1dOT/CIiEgklGBERiYQSjIiIREIJRkREIqEEIyIikejSd5GZWTWwtr3jaKVsggdLO4qOFi8o5lOho8ULHS/mKOM9zd3jPqke2XMwHcTaRG61SyZmVtKRYu5o8YJiPhU6WrzQ8WJOhnh1iUxERCKhBCMiIpHo6gnmgfYO4AR0tJg7WrygmE+FjhYvdLyY2z3eLt3JLyIi0enqZzAiIhKRDp9gzOwhM6s0sxUxZZlm9rqZrQ9/DwjLzcx+bWalZrbMzApjtrk5rL/ezG6OKT/HzJaH2/w6nJ45iph/bmZrwrjmmFlGzLq7wvdfa2ZXx5RPCctKzezOmPIRZvZBuC9PhlMbtHnMMev+0czczLLD1+1+nJuL18z+PjxmK83sZzHlSXmMzexsM3vfzJaEE+VNDsuT4RgPM7O3zGx1eDy/G5Yn7eevhZiT8vPXXLwx65Pus3ccd+/QP8AlQCGwIqbsZ8Cd4fKdwL3h8lTgFYIZMM8HPgjLM4EN4e8B4fKAcN184IJwm1eAayKK+TNA93D53piYxwFLgVRgBPAhwVQFKeHySKBnWGdcuM1TwKxw+ffAN6KIOSwfRjB9wmYgO1mOczPH+DLgDSA1fD0o2Y8x8FrDsQiP65+T6BgPBQrD5XRgXXgsk/bz10LMSfn5ay7eZP7sxf50+DMYd3+HYC6ZWNOBh8Plh4HrY8of8cD7BLNgDgWuBl539yp33wO8DkwJ1/Vz93ke/Es8EtNWm8bs7q+5e2348n2
```
pllim: Please remove the `# photon s-1 cm^-2 AA^-1`, as per the other example above.

tutorials/observing_examples.ipynb
bmorris3: It may be naive but I'm not sure we can do much better for ground-based observations without sky brightness models, which are complicated (this is something we've come upon in [astroplan again and again](https://github.com/astropy/astroplan/issues/29)).

docs/index.rst
tcjansen: Ah, yeah, that's just me having some git confusion with fetching and branches...

tutorials/observing_examples.ipynb
bmorris3: I think the factor of two is pretty good, given that (1) we're ignoring the atmosphere, and (2) we're assuming a blackbody.

bmorris3:
Copying this message from slack for continuity: 

So perhaps the right thing to do is to see if we can do better than a factor of two (for the count rate of HAT-P-11) by sprinting a little on this particular problem.

telescopy has atmospheric transmittance curves which we can assume are generic for now to see if they get us closer to the true count rate (they’re actually model transmittance curves for Cerro Paranal, in particular). You can find the two transmittance curves [here](https://github.com/bmorris3/telescopy/tree/master/telescopy/data), one for X=1 (airmass) and another for X=1.5. The full references and API for using this model within telescopy is [here](https://github.com/bmorris3/telescopy/blob/master/telescopy/skymodel.py). Let’s use X=1.5 for now, because it’s closer to the true observation. The file has two columns: wavelength in nm, and transmittance. You should be able to create a `SpectralElement` in synphot with this X=1.5 transmittance curve, and then create a composite `SpectralElement` with both the SDSS filter and the transmittance curve by multiplying together each `SpectralElement` with the `*` operator. Then recompute the counts and see if they move in the right direction – if you’re applying the new curve correctly, it should _decrease_ the count rate for both stars.

tcjansen:
Copying from slack:

Okay, I've added the atmospheric transmittance model (with airmass=1.5), and the counts are indeed lower, which is the good news! The sort-of band news is that it didn't reduce by the factor of 2 we were hoping for (old counts were = 67.5 million, new counts = 58.5 million, actual counts = 34.5 million), but similarly to what @bmorris3 mentioned in slack this could just be due to differences in airmass between the time of observation and our 1.5 airmass model.

bmorris3:
Latest precision recap at commit 0c94d79: 
```
synphot T-1: 261K
actual T-1: 203K
synphot H-11: 30M
actual H-11: 34M
```

pllim:
@tcjansen , just want to let you know that I am very happy with the progress you made using `synphot` in your workflow in a week. Keep up the good work! 👏

tcjansen:
Thanks @pllim, I appreciate that!

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,587 @@
...
+   "metadata": {},
+   "source": [
+    "### Simulate the observation with a specific instrument"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<p>Let's observe HAT-P-11 with the Agile instrument on APO's 3.5m telescope, whose net wavelength transmission can be downloaded from <a href=\"https://www.apo.nmsu.edu/arc35m/Instruments/AGILE/Anjum/wav.dat\">APO's website</a>. We will use this to construct a bandpass specific to Agile.</p><p>To construct a bandpass from a file with <tt>synphot</tt>, use <a href=\"https://synphot.readthedocs.io/en/latest/api/synphot.spectrum.SpectralElement.html#synphot.spectrum\">SpectralElement</a>:</p>"
```
bmorris3: The Agile exercise was great, but now that we're closing in on reproducing the ARCTIC observations, let's get rid of the Agile example.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,587 @@
...
...ROsbZA3O1DdqDgD1YCqna44VWeghVcMas7ANKMVJqrOJUpWc1RhvHO7ZOiGM6gm+DVFqxVWKIkEnrWxnEFaT9NvqcxofhXcg1MPAnDT2E2rthWKgrZjDGdqddVQOzE4OneU3UX/u6tPRFAMpJQO8Bbgq8APgM9IKR8TQvy+EOLO6mZfBeaEEEeBbwD/l5Rydb/2OlgsWwy1WeGsGetaAtMqgJYK9ygKaBhHEYXS0qCaiLIZqTBRMLE0mak6mzhn0DM24wws10JTNKRtI3R/kafge+E7g9Vi0GnnftqEp7w5bNfmyNQRdmd3szPTfK0Tfec49uTUqueHkkOMpkY5vnB81WtH545y0B4GVUXbEb7VFTta+FdK+WXgyyuee3fd3yXw69U/XWGhaHPVjvV12maTQQK5E2dQDH9ZKSyLgZSREINapZNpRitMFIhBtioGcTVR79ikMwjCRELXSahKLUykJBJNE8hlp4wr3bbOQNo2uumwpCt869y3uPf8vbzm6te03F4b34n5g8ebvnbD6A18b+Z7Dc+Zrsmx+WO8sbgLbecOhBq+m9DQFoEvliwG1+kMMtX1kAud5gzCHiICP0yEBM+NRDVR4AxExYpYmMhGMxS0VLVePXYGvaNaWrpeZyClxPGc2qA6YRgktWQtTCSSyaYJ5E7WDQ+WafUySf7rN/8rlmfxqitf1XJ7fec49tRU06qhm3bcxJncGebKy8GRh6cfxvIsdsy7GJeFcwROKMXAcjyKlsvQGusfr6Qv4V+IOhpjbRbCnzwG3xkAeHYtZxDmPoMggSwqVqTCRGbRJpnREUbgDGIx6AW17mNYtzOwq82X/gjrwBkklhPICaNpArndKIoAL+9v8+LrX8lAYoCfvfZnuW7kupbba+M7keUyXm712hc377gZ8AUg4NsXvo0mNBIX5jH2h1MMOgoTbTULJf+LuN6cQVJXUMQ6Sksj4wwA16pVE0WhA1kx7YiFiRwSGR2hVxPgcZioN9SJwXr7DGpioNSLgVpzzEoiiZtfvUpdJ2IQvO+mA8/jnhe9u+V2Afq4P4fInpxCHWgcLXH9yPVk9Sz/ev5fedG+FwFw74V7OZy+Fm/pIYx9+9fc/3YQSmfw1Lyf0d8ztL47SyEE2YTWedNZFJxBkJB1bSpuBV3RUUQof21A1bVIiWLZ0QoTFVY4gzhM1BM25QyqzZeao
```
bmorris3: Like in my comment above, let's keep the demo notebook just focused on ARCTIC for now.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,587 @@
...
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "T_eff = 4780  # temperature of hatp11[K], Bakos et al. 2010\n",
+    "\n",
+    "# Used the phoenix module in aesop.py to generate a model spectrum for T_eff = 4780\n",
+    "photlam_hatp11 = SourceSpectrum.from_file('phoenix_model_spectrum.dat', flux_unit = u.erg / u.s / u.cm ** 2 / u.cm)\n",
```
bmorris3: Let's do some astropy magic here rather than including a data file in the PR. You can [construct a `SourceSpectrum` from a wavelength and flux array like so](https://synphot.readthedocs.io/en/latest/synphot/spectrum.html#arrays), and you can use astropy's FITS/downloading machinery like I did [in this notebook](https://gist.github.com/bmorris3/d661dd739c5c0f44f61daf2b9cca1d2a) in order to get the wavelength and flux arrays.

tutorials/agile.dat
```diff
@@ -0,0 +1,69 @@
+320 0.0561637
```
bmorris3: When removing the Agile exercise, you can remove this file too.

bmorris3:
After you set up algorithmic spectra retrieval for HAT-P-11, do the same for the PHOENIX spectrum of an M8V (TRAPPIST-1): ftp://phoenix.astro.physik.uni-goettingen.de/v2.0/HiResFITS/PHOENIX-ACES-AGSS-COND-2011/Z-0.0/lte02500-5.00-0.0.PHOENIX-ACES-AGSS-COND-2011-HiRes.fits

bmorris3:
Another enhancement proposal (feel free to poke holes in the idea!): since we'll have PHOENIX model spectra of HAT-P-11 and TRAPPIST-1, it would be fun to model their count rates through a second telescope. I'm thinking perhaps we can use Kepler, since (1) Kepler observed both HAT-P-11 (Kepler) and TRAPPIST-1 (K2); (2) it's a space telescope and therefore shouldn't depend on our atmosphere model, making it a good sanity check. 

Here's some of the info you'll need to do Kepler: 
* [Kepler instrument+filter response](https://keplergo.arc.nasa.gov/kepler_response_hires1.txt)
* Answer key [via MAST](https://mast.stsci.edu/portal/Mashup/Clients/Mast/Portal.html): HAT-P-11 measured counts: 2.8e6 e-/s; TRAPPIST-1 measured counts: 5.70e3 e-/s
* Kepler Gain (which you might not need since we're reporting e-/s above): 110 e-/ADU [source](https://archive.stsci.edu/kepler/manuals/KSCI-19033-001.pdf)

bmorris3:
Another enhancement proposal: use astroquery to get the stellar distance and radius like I do in [this example](https://gist.github.com/bmorris3/685db348daf6dac409e2ee1726f0a60b), rather than hard-coding the answers into the notebook. You'll need to come up with a more clever way of selecting the correct answer from the query response than choosing the first row, as I did in the example. You might find that the answer is the star closest to the input coordinates. You can measure distances between coordinates using [`SkyCoord.separation()`](http://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.separation).

bmorris3:
Enhancement proposal: ultimately you'll be making a PR to astroquery which queries the [VO service from the SVO Filter Profile Service](http://svo2.cab.inta-csic.es/theory/fps/index.php?mode=voservice), but for now, you can do something simpler to retrieve the filter response curves for Sloan filters. Use the `astropy.utils.data.download_file` function to grab the response curves for each filter from the following links (and remove the step in the tutorial which says 'you can download the filter response curves here...').
* [SDSS r](http://svo2.cab.inta-csic.es/theory/fps/getdata.php?format=ascii&id=SLOAN/SDSS.r)
* [SDSS z](http://svo2.cab.inta-csic.es/theory/fps/getdata.php?format=ascii&id=SLOAN/SDSS.z)

tcjansen:
> After you set up algorithmic spectra retrieval for HAT-P-11 ([comment](https://github.com/spacetelescope/synphot_refactor/pull/192#discussion_r289672858)), do the same for the PHOENIX spectrum of an M8V (TRAPPIST-1)

Okay, as of now I've made it as far as the above suggestion! The results using the PHOENIX spectra, QE, and Cerro Paranal transmission at 1.5 atmos are:
```
synphot TRAPPIST-1: 257K
actual TRAPPIST-1: 203K
synphot HAT-P-11: 30M
actual HAT-P-11: 34M
```
I'll get working on the other suggestions ASAP!

docs/index.rst
bmorris3: Still need to get rid of the changes to this file in this PR. Lemme know if you need help with that!

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,441 @@
...
+    }
+   ],
+   "source": [
+    "photlam_hatp11 = SourceSpectrum(Empirical1D,\n",
+    "                                points=wavelengths, lookup_table=flux)\n",
+    "\n",
+    "# Compute the flux at Earth from the luminosity \n",
+    "R = 0.683 * u.R_sun  # radius in solar units, Deming et al. 2011\n",
+    "d = 123 * u.lyr # distance to HATP11 in light years from Gaia parallax\n",
+    "photlam_hatp11 = photlam_hatp11 * float(R / d) ** 2 / np.pi\n",
```
bmorris3: If/when we figure out this extra factor of pi, we should leave a comment explaining why it's necessary.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,441 @@
...
+    "<p>Let's observe HAT-P-11 with the <a href=\"https://www.apo.nmsu.edu/arc35m/Instruments/ARCTIC/\">ARCTIC instrument</a> on APO's 3.5m telescope."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 131,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "r = 3.5 / 2 * u.m  # radius of 3.5m ARC telescope at APO\n",
```
bmorris3: In my next few comments, I hope you'll forgive me as I get a bit nitpicky about variable names. Zen of Python tells us
```
Explicit is better than implicit...
Readability counts.
```
In that spirit, and in keeping with the [astropy-tutorials examples](https://github.com/astropy/astropy-tutorials/blob/master/tutorials/notebooks/coordinates/coordinates.ipynb), let's give variables descriptive, full-word names. So:
```
r -> aperture_radius
R -> stellar_radius
```
etc.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,441 @@
...
...ePo1mSJg/zqw6Fub68ik8F6f3Ncsgrc77mtXnruO6U3hzVqdDrcALrypE9vA4hJeVmZ6ZFH6WGxpX65LaTfdHMvqn1P7e8OM+hSJyjSaSJ1m7fx+2vzKe8a2t+fJKeBJuiR0lLr0NQHvDjoIJ+tW5HpdchHEaTSBMYY7jxubmEQob7vzswUHM1K//oVpze/Rrcap01qIvO+5IMWifSBC/OWsu0FVv4zblH0blNc6/DUQHVqXUeKzfv8TqMlJPdQD3Q45cOYdveA0mMJnVpEknQjr1V/GbyIo7p0orvlnf2OhylVBy0D5dzNIkk6LdvLmbb3gM8cdmxaVFpqdzToTCv3tfOCNjMe37UqXX9f1/VdFonkoC5a7bz5Gerufj4Mvp1SKwnsVK1MsL+C4tb1h2na5AObd5kvz1vgNchHFSSgpNWaRKJU03IcPsr8ylumcN1OkOhckB4B7LImfH6pkOPc6fq1espEPBTv99ebfN9FY8TtDgrTk9/vpp5FTt44PyBFORqp0LljteuPYGqGsPAztrCKNUc1bGQeRU7vA7DMZpE4rB5937ue2Mxw7oXcZYO0aFclCHCwM7acVX5nxZnxWHi64vZV1XD3ef017GxlFIKTSIxm75qKy/MrODyEd3pWZoG5dQq8P5xUbnXISTVYZd1PhysMBVpEolBdU2I21+eT8dWeVwzuqfX4SgVk9L81GsJ1JD6cobfygz8OBJvU2gSicHjn6xi8fpd3HFmvyYPya2UUqlEz4iN2LCzkvvf+pJRR5Qwtl9br8NRHnn1mhOYvWa712HEpX+a9WHy4o7jzAEd+N/cr+N6T7KrU6ct38KabXtd276rdyIiMk5ElojIMhG5tZ51viMiC0VkgYg85WY8iXjgnaUcqAlx51lamZ7OjuxYyPeP6+r6fpws6shK8zlEkuHP4wd5HUKjxv/jU25+wb0h5F27ExGRTOCvwClABTBdRCYZYxaGrdMLuA0YbozZJiK+GtBm9Za9PDd9Dd87tgtdi9J7pFXlnmjDcuj1igoKNy9VhgLLjDErjDEHgGeAsyPW+SHwV2PMNgBjzEYX44nbA+8sJTNDuGqUVqYr91w5Ur9f6eiEnsVeh+AIN5NIR2BN2PMKe1m43kBvEflYRD4VkXEuxhOXZRt38dLsCi4a1pW2BTqtpnKPG/PQvHrNCY5v0y2OjXqit2+ecD
```
bmorris3:
```
qe -> quantum_efficiency
```

bmorris3: Let's print the `int` of the count here.

bmorris3: See above comment RE: `int`.

tutorials/observing_examples.ipynb
```diff
@@ -0,0 +1,441 @@
...
+    "%matplotlib inline\n",
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np\n",
+    "\n",
+    "import astropy.units as u\n",
+    "from astropy.io import fits\n",
+    "from astropy.modeling.blackbody import blackbody_lambda\n",
+    "\n",
+    "import synphot.units as units\n",
+    "from synphot.models import BlackBody1D, BlackBodyNorm1D, Empirical1D\n",
```
bmorris3: Let's remove any imports we don't need anymore, like `BlackBodyNorm1D`, `blackbody_lambda`

tutorials/observing_examples.ipynb
bmorris3: Let's make a final cell which compares the synphot expected count rate (y) with the measured count rate (x), with a 1:1 line over plotted.

tutorials/observing_examples.ipynb
tcjansen: Good to know - I prefer being more explicit but was worried about being too "wordy"!

tutorials/observing_examples.ipynb
tcjansen: Do you have a paper I could cite for the observed values?

tcjansen:
I'm moving changes to my own branched repo [here](https://github.com/tcjansen/GSoC/tree/the-first-notebook)

pllim:
Great! @bmorris3 and @tcjansen , once you are done moving all the comments and materials over to the other repo, please close this PR.

tutorials/observing_examples.ipynb
bmorris3: I'm definitely pro-"wordy". The closer the code reads to an english sentence the better!

tutorials/observing_examples.ipynb
bmorris3: The observed values aren't published in papers, but the observations themselves make up two research notes: [HAT-P-11](http://adsabs.harvard.edu/abs/2018RNAAS...2a..26M) and [TRAPPIST-1](http://adsabs.harvard.edu/abs/2018RNAAS...2a..10M).

pllim:
Looks like things are moving along smoothly over there, so I am closing this one.

PR #199, opened 2019-06-14
TST: Add [docs only] directive

### LINKED ISSUE #196: TST: Actions to recognize [docs only] in commit message
~Azure~ Actions to recognize `[docs only]` directive in commit message and then only run the docs job and skip everything else, except PEP 8 check, which should always run regardless.

pllim:
Inspired by https://github.com/Microsoft/azure-pipelines-agent/issues/858#issuecomment-457027046 . Let's see if this works. Only the "initial" jobs and doc build should run, nothing else.

Attempt to fix #196


pllim:
Grrr, doesn't work. Battle for another day.

PR #204, opened 2019-07-10
Adding signal to noise ratio calculations to observation.py

tcjansen:
I have added two functions to `observation.py`:

- `howell_snr()` which returns the idealized theoretical signal to noise ratio from a given number of counts. The expression is taken from pg 55 of "Handbook of CCD Astronomy", Steve B. Howell, 2000. 
- `exptime_from_howell_snr()` which returns the exposure time needed in units of seconds to achieve the specified (idealized theoretical) signal to noise ratio (from pg 57-58 of Howell 2000).

Some tests for these functions have also been added to `tests/test_observation.py`

### DIFF
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -726,7 +726,7 @@ def ccd_snr(counts,
         Total number of counts in an arbitrary exposure time with units of
         either astropy.units.ct, astropy.units.electron, or
         astropy.units.photon. Be aware that count units may refer to different
-        physical units for different instruments (e.g. one instrument may be
+        physical units for different instruments (e.g., one instrument may be
         "count"ing electrons, while another may be counting analog-to-digital
         units [ADU]). In the latter case, the user must first multiply their
         counts by the instrument gain to reflect the correct number of counts
@@ -782,7 +782,7 @@ def ccd_snr(counts,
         counts = counts.value * u.ct
     elif counts.unit != u.ct:
         raise u.UnitsError('counts must have units of either '
-                           'astropy.units.ct, astropy.units.electron '
+                           'astropy.units.ct, astropy.units.electron, '
                            'or astropy.units.photon')
 
     readnoise = _get_shotnoise(readnoise)
@@ -880,10 +880,10 @@ def exptime_from_ccd_snr(snr, countrate,
         to noise ratio.
     """
     if countrate.unit in (u.electron / u.s, u.photon / u.s):
-        countrate = countrate.value * u.ct / u.s
+        countrate = countrate.value * (u.ct / u.s)
     elif countrate.unit != u.ct / u.s:
         raise u.UnitsError('countrate must have units of (either '
-                           'astropy.units.ct, astropy.units.electron or '
+                           'astropy.units.ct, astropy.units.electron, or '
                            'astropy.units.photon) / astropy.units.s')
 
     # necessary for units to work:
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -918,7 +918,7 @@ def _get_shotnoise(detector_property):
     """
     # Ensure detector_property is in the correct units:
     detector_property = detector_property.to(u.ct / u.pixel)
-    return detector_property.value * np.sqrt(1 * u.ct / u.pixel)
+    return detector_property.value * np.sqrt(1 * (u.ct / u.pixel))
 
 
 def _t_with_small_errs(t, background_rate, darkcurrent_rate, gain_err,
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -22,7 +22,7 @@ from . import binning, exceptions, units, utils
 from .models import Empirical1D
 from .spectrum import BaseSourceSpectrum, SourceSpectrum, SpectralElement
 
-__all__ = ['Observation']
+__all__ = ['Observation', 'howell_snr', 'exptime_from_howell_snr']
 
 
 class Observation(BaseSourceSpectrum):
@@ -715,7 +715,7 @@ def howell_snr(counts,
                darkcurrent=0 * (u.electron / u.pixel),
                readnoise=0 * (u.electron / u.pixel),
                gain=1 * (u.electron / u.adu),
-               ad_err=ad_err_default):
+               ad_err=AD_ERR_DEFAULT):
     """
     A function to calculate the idealized theoretical signal to noise ratio
     (SNR) of an astronomical observation with a given number of counts. This
@@ -801,7 +801,7 @@ def exptime_from_howell_snr(snr, countrate,
                             darkcurrent_rate=0 * (u.electron / u.pixel / u.s),
                             readnoise=0 * (u.electron / u.pixel),
                             gain=1 * (u.electron / u.adu),
-                            ad_err=ad_err_default):
+                            ad_err=AD_ERR_DEFAULT):
     """
     Returns the exposure time needed in units of seconds to achieve
     the specified (idealized theoretical) signal to noise ratio
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -10,7 +10,6 @@ import warnings
 
 # THIRD-PARTY
 import numpy as np
-from scipy.optimize import fsolve
 
 # ASTROPY
 from astropy import log
@@ -697,7 +696,8 @@ class Observation(BaseSourceSpectrum):
                               meta={'header': header})
 
 
-ad_err_default = np.sqrt(0.289) * u.adu / u.pixel
+# See ad_err documentation below.
+AD_ERR_DEFAULT = np.sqrt(0.289) * (u.adu / u.pixel)
 
 
 @u.quantity_input(counts=u.electron,
@@ -711,10 +711,10 @@ ad_err_default = np.sqrt(0.289) * u.adu / u.pixel
 def howell_snr(counts,
                npix=1 * u.pixel,
                n_background=np.inf * u.pixel,
-               background=0 * u.electron / u.pixel,
-               darkcurrent=0 * u.electron / u.pixel,
-               readnoise=0 * u.electron / u.pixel,
-               gain=1 * u.electron / u.adu,
+               background=0 * (u.electron / u.pixel),
+               darkcurrent=0 * (u.electron / u.pixel),
+               readnoise=0 * (u.electron / u.pixel),
+               gain=1 * (u.electron / u.adu),
                ad_err=ad_err_default):
     """
     A function to calculate the idealized theoretical signal to noise ratio
@@ -797,10 +797,10 @@ def howell_snr(counts,
 def exptime_from_howell_snr(snr, countrate,
                             npix=1 * u.pixel,
                             n_background=np.inf * u.pixel,
-                            background_rate=0 * u.electron / u.pixel / u.s,
-                            darkcurrent_rate=0 * u.electron / u.pixel / u.s,
-                            readnoise=0 * u.electron / u.pixel,
-                            gain=1 * u.electron / u.adu,
+                            background_rate=0 * (u.electron / u.pixel / u.s),
+                            darkcurrent_rate=0 * (u.electron / u.pixel / u.s),
+                            readnoise=0 * (u.electron / u.pixel),
+                            gain=1 * (u.electron / u.adu),
                             ad_err=ad_err_default):
     """
     Returns the exposure time needed in units of seconds to achieve
@@ -826,12 +826,12 @@ def exptime_from_howell_snr(snr, countrate,
         Photons per pixel per second due to the backround/sky with units of
         electrons/second/pixel.
         Default is 0 * (astropy.units.electron /
-                        astropy.units.second / astropy.units.pixel)
+        astropy.units.second / astropy.units.pixel)
     darkcurrent_rate : `~astropy.units.Quantity`, optional
         Electrons per pixel per second due to the dark current with units
         of electrons/second/pixel.
         Default is 0 * (astropy.units.electron /
-                        astropy.units.second / astropy.units.pixel)
+        astropy.units.second / astropy.units.pixel)
     readnoise : `~astropy.units.Quantity`, optional
         Electrons per pixel from the read noise with units of electrons/pixel.
         Default is 0 * astropy.units.electron / astropy.units.pixel.
@@ -875,6 +875,7 @@ def exptime_from_howell_snr(snr, countrate,
     t = (-B + np.sqrt(B ** 2 - 4 * A * C)) / (2 * A)
 
     if gain_err.value > 1 or not np.isfinite(n_background.value):
+        from scipy.optimize import fsolve
         # solve t numerically
         t = fsolve(_t_with_small_errs, t, args=(background_rate,
                                                 darkcurrent_rate,
@@ -888,7 +889,7 @@ def exptime_from_howell_snr(snr, countrate,
 def _get_shotnoise(detector_property):
     """
     Returns the shot noise (i.e. non-Poissonion noise) in the correct
-    units.
+    units. ``detector_property`` must be a Quantity.
     """
     # Ensure detector_property is in the correct units:
     detector_property = detector_property.to(u.electron / u.pixel)
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -773,8 +773,8 @@ def howell_snr(counts,
     Returns
     -------
     snr : `~astropy.units.Quantity`
-        The signal to noise ratio of the given observation in units of
-        sqrt(counts).
+        The signal to noise ratio of the given observation in dimensionless
+        units.
     """
     readnoise = _get_shotnoise(readnoise)
     gain_err = _get_shotnoise(gain * ad_err)
@@ -783,11 +783,11 @@ def howell_snr(counts,
     detector_noise = (background + darkcurrent +
                       readnoise ** 2 + gain_err ** 2)
 
-    return counts / np.sqrt(counts + pixel_terms * detector_noise)
+    return (counts / np.sqrt(counts + pixel_terms * detector_noise) / 
+            np.sqrt(1 * u.ct))
 
 
-@u.quantity_input(snr=np.sqrt(1 * u.ct),
-                  countrate=u.ct / u.s,
+@u.quantity_input(countrate=u.ct / u.s,
                   npix=u.pixel,
                   n_background=u.pixel,
                   background_rate=u.ct / u.pixel / u.s,
@@ -810,9 +810,9 @@ def exptime_from_howell_snr(snr, countrate,
 
     Parameters
     ----------
-    snr : `~astropy.units.Quantity`
-        The signal to noise ratio of the given observation in units of
-        sqrt(counts).
+    snr : float, int, or `~astropy.units.Quantity`
+        The signal to noise ratio of the given observation in dimensionless
+        units.
     countrate : `~astropy.units.Quantity`
         The counts per second with units of astropy.units.ct/second. Be aware
         that the count unit may refer to different physical units for different
@@ -867,6 +867,8 @@ def exptime_from_howell_snr(snr, countrate,
         The exposure time needed (in seconds) to achieve the given signal
         to noise ratio.
     """
+    # necessary for units to work:
+    snr = snr * np.sqrt(1 * u.ct)
     readnoise = _get_shotnoise(readnoise)
     gain_err = _get_shotnoise(gain * ad_err)
 
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -22,7 +22,7 @@ from . import binning, exceptions, units, utils
 from .models import Empirical1D
 from .spectrum import BaseSourceSpectrum, SourceSpectrum, SpectralElement
 
-__all__ = ['Observation', 'howell_snr', 'exptime_from_howell_snr']
+__all__ = ['Observation', 'ccd_snr', 'exptime_from_ccd_snr']
 
 
 class Observation(BaseSourceSpectrum):
@@ -700,22 +700,21 @@ class Observation(BaseSourceSpectrum):
 AD_ERR_DEFAULT = np.sqrt(0.289) * (u.adu / u.pixel)
 
 
-@u.quantity_input(counts=u.ct,
-                  npix=u.pixel,
+@u.quantity_input(npix=u.pixel,
                   n_background=u.pixel,
                   background=u.ct / u.pixel,
                   darkcurrent=u.ct / u.pixel,
                   readnoise=u.ct / u.pixel,
                   gain=u.ct / u.adu,
                   ad_err=u.adu / u.pixel)
-def howell_snr(counts,
-               npix=1 * u.pixel,
-               n_background=np.inf * u.pixel,
-               background=0 * (u.ct / u.pixel),
-               darkcurrent=0 * (u.ct / u.pixel),
-               readnoise=0 * (u.ct / u.pixel),
-               gain=1 * (u.ct / u.adu),
-               ad_err=AD_ERR_DEFAULT):
+def ccd_snr(counts,
+            npix=1 * u.pixel,
+            n_background=np.inf * u.pixel,
+            background=0 * (u.ct / u.pixel),
+            darkcurrent=0 * (u.ct / u.pixel),
+            readnoise=0 * (u.ct / u.pixel),
+            gain=1 * (u.ct / u.adu),
+            ad_err=AD_ERR_DEFAULT):
     """
     A function to calculate the idealized theoretical signal to noise ratio
     (SNR) of an astronomical observation with a given number of counts. This
@@ -776,6 +775,13 @@ def howell_snr(counts,
         The signal to noise ratio of the given observation in dimensionless
         units.
     """
+    if counts.unit in (u.electron, u.photon):
+        counts = counts.value * u.ct
+    elif counts.unit != u.ct:
+        raise u.UnitsError('counts must have units of either '
+                           'astropy.units.ct, astropy.units.electron '
+                           'or astropy.units.photon')
+
     readnoise = _get_shotnoise(readnoise)
     gain_err = _get_shotnoise(gain * ad_err)
 
@@ -787,22 +793,21 @@ def howell_snr(counts,
             np.sqrt(1 * u.ct))
 
 
-@u.quantity_input(countrate=u.ct / u.s,
-                  npix=u.pixel,
+@u.quantity_input(npix=u.pixel,
                   n_background=u.pixel,
                   background_rate=u.ct / u.pixel / u.s,
                   darkcurrent_rate=u.ct / u.pixel / u.s,
                   readnoise=u.ct / u.pixel,
                   gain=u.ct / u.adu,
                   ad_err=u.adu / u.pixel)
-def exptime_from_howell_snr(snr, countrate,
-                            npix=1 * u.pixel,
-                            n_background=np.inf * u.pixel,
-                            background_rate=0 * (u.ct / u.pixel / u.s),
-                            darkcurrent_rate=0 * (u.ct / u.pixel / u.s),
-                            readnoise=0 * (u.ct / u.pixel),
-                            gain=1 * (u.ct / u.adu),
-                            ad_err=AD_ERR_DEFAULT):
+def exptime_from_ccd_snr(snr, countrate,
+                         npix=1 * u.pixel,
+                         n_background=np.inf * u.pixel,
+                         background_rate=0 * (u.ct / u.pixel / u.s),
+                         darkcurrent_rate=0 * (u.ct / u.pixel / u.s),
+                         readnoise=0 * (u.ct / u.pixel),
+                         gain=1 * (u.ct / u.adu),
+                         ad_err=AD_ERR_DEFAULT):
     """
     Returns the exposure time needed in units of seconds to achieve
     the specified (idealized theoretical) signal to noise ratio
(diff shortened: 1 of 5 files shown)


pllim:
cc @bmorris3

pllim:
Please fix PEP 8 warnings:
```
synphot/observation.py:18:1: F401 'astropy.tests.helper.assert_quantity_allclose' imported but unused
synphot/observation.py:910:49: E226 missing whitespace around arithmetic operator
```

synphot/observation.py
```diff
@@ -10,6 +10,7 @@
 
 # THIRD-PARTY
 import numpy as np
+from scipy.optimize import fsolve
```
pllim:
```suggestion
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+    # solve t with the quadratic equation (pg. 57 of Howell 2000)
+    A = countrate ** 2
+    B = (-1) * snr ** 2 * (countrate + npix * (background_rate +
+                                               darkcurrent_rate))
+    C = (-1) * snr ** 2 * npix * readnoise ** 2
+
+    t = (-B + np.sqrt(B ** 2 - 4 * A * C)) / (2 * A)
+
+    if gain_err.value > 1 or not np.isfinite(n_background.value):
+        # solve t numerically
```
pllim: Since `fsolve` is only used for a very specific case, I would prefer not to import it until it is absolutely needed.
```suggestion
        from scipy.optimize import fsolve
        # solve t numerically
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
         header = {'observation': str(self), 'binned': binned}
         return SourceSpectrum(Empirical1D, points=w, lookup_table=y,
                               meta={'header': header})
+
+
+ad_err_default = np.sqrt(0.289) * u.adu / u.pixel
```
pllim: Change variable to `AD_ERR_DEFAULT` to indicate that it is a constant.
```suggestion
# See ad_err documentation below.
AD_ERR_DEFAULT = np.sqrt(0.289) * (u.adu / u.pixel)
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+                  n_background=u.pixel,
+                  background=u.electron / u.pixel,
+                  darkcurrent=u.electron / u.pixel,
+                  readnoise=u.electron / u.pixel,
+                  gain=u.electron / u.adu,
+                  ad_err=u.adu / u.pixel)
+def howell_snr(counts,
+               npix=1 * u.pixel,
+               n_background=np.inf * u.pixel,
+               background=0 * u.electron / u.pixel,
```
pllim:
```suggestion
               background=0 * (u.electron / u.pixel),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+               darkcurrent=0 * u.electron / u.pixel,
```
pllim:
```suggestion
               darkcurrent=0 * (u.electron / u.pixel),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+               readnoise=0 * u.electron / u.pixel,
```
pllim:
```suggestion
               readnoise=0 * (u.electron / u.pixel),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+               gain=1 * u.electron / u.adu,
```
pllim:
```suggestion
               gain=1 * (u.electron / u.adu),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+                  n_background=u.pixel,
+                  background_rate=u.electron / u.pixel / u.s,
+                  darkcurrent_rate=u.electron / u.pixel / u.s,
+                  readnoise=u.electron / u.pixel,
+                  gain=u.electron / u.adu,
+                  ad_err=u.adu / u.pixel)
+def exptime_from_howell_snr(snr, countrate,
+                            npix=1 * u.pixel,
+                            n_background=np.inf * u.pixel,
+                            background_rate=0 * u.electron / u.pixel / u.s,
```
pllim:
```suggestion
                            background_rate=0 * (u.electron / u.pixel / u.s),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+                            darkcurrent_rate=0 * u.electron / u.pixel / u.s,
```
pllim:
```suggestion
                            darkcurrent_rate=0 * (u.electron / u.pixel / u.s),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+                            readnoise=0 * u.electron / u.pixel,
```
pllim:
```suggestion
                            readnoise=0 * (u.electron / u.pixel),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+                            gain=1 * u.electron / u.adu,
```
pllim:
```suggestion
                            gain=1 * (u.electron / u.adu),
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+    n_background : `~astropy.units.Quantity`, optional
+        Number of pixels used in the background estimation with units of
+        pixels. Default is set to np.inf * astropy.units.pixel
+        such that there is no contribution of error due to background
+        estimation. This assumes that n_background will be >> npix.
+    background_rate : `~astropy.units.Quantity`, optional
+        Photons per pixel per second due to the backround/sky with units of
+        electrons/second/pixel.
+        Default is 0 * (astropy.units.electron /
+                        astropy.units.second / astropy.units.pixel)
```
pllim: No need for so much indentation in a docstring.
```suggestion
        astropy.units.second / astropy.units.pixel)
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+    darkcurrent_rate : `~astropy.units.Quantity`, optional
+        Electrons per pixel per second due to the dark current with units
+        of electrons/second/pixel.
+        Default is 0 * (astropy.units.electron /
+                        astropy.units.second / astropy.units.pixel)
```
pllim:
```suggestion
        astropy.units.second / astropy.units.pixel)
```

synphot/observation.py
```diff
@@ -694,3 +695,218 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+                                                npix, n_background))
+        t = t * u.s
+
+    return t
+
+
+def _get_shotnoise(detector_property):
+    """
+    Returns the shot noise (i.e. non-Poissonion noise) in the correct
+    units.
```
pllim:
```suggestion
    units. ``detector_property`` must be a Quantity.
```

synphot/tests/test_observation.py
```diff
@@ -22,6 +23,8 @@
 from ..models import (BlackBodyNorm1D, Box1D, ConstFlux1D, Empirical1D,
                       GaussianFlux1D)
 from ..observation import Observation
```
pllim: You can import them all in one line. (Eh, not sure why GitHub diff is showing the removal of a leading whitespace; there should be none.)
```suggestion
from ..observation import Observation, howell_snr, exptime_from_howell_snr
```

synphot/tests/test_observation.py
```diff
@@ -22,6 +23,8 @@
...
+from ..observation import howell_snr
```
pllim:
```suggestion
```

synphot/tests/test_observation.py
```diff
@@ -22,6 +23,8 @@
...
+from ..observation import exptime_from_howell_snr
```
pllim:
```suggestion
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    n_background = 200 * u.pixel
+    npix = 1 * u.pixel
+    counts = 24013 * u.adu * gain
+
+    result = howell_snr(counts, npix=npix, background=background,
+                        darkcurrent=darkcurrent, readnoise=readnoise,
+                        gain=gain, n_background=n_background)
+
+    # the value of the answer given in the text
+    answer = 342 * np.sqrt(1 * u.electron)
```
pllim:
```suggestion
    ans_unit = np.sqrt(1 * u.electron)
    answer = 342 * ans_unit
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+
+    # allow error to be +/- 1 for this test
+    assert_quantity_allclose(result, answer, atol=1 * np.sqrt(1 * u.electron))
```
pllim:
```suggestion
    assert_quantity_allclose(result, answer, atol=1 * ans_unit)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    A test to check that the math in observation.howell_snr is done correctly.
+    Based on the worked example on pg. 56-57 of [1]_.
+
+    References
+    ----------
+    .. [1] Howell, S. B. 2000, *Handbook of CCD Astronomy* (Cambridge, UK:
+        Cambridge University Press)
+    """
+    t = 300 * u.s
+    readnoise = 5 * u.electron / u.pixel
```
pllim:
```suggestion
    readnoise = 5 * (u.electron / u.pixel)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    darkcurrent = 22 * t * u.electron / u.pixel / u.hr
```
pllim:
```suggestion
    darkcurrent = 22 * t * (u.electron / u.pixel / u.hr)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    gain = 5 * u.electron / u.adu
```
pllim:
```suggestion
    gain = 5 * (u.electron / u.adu)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    background = 620 * u.adu / u.pixel * gain
```
pllim:
```suggestion
    background = 620 * (u.adu / u.pixel * gain)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    example on pg. 56-57 of [1]_.
+
+    References
+    ----------
+    .. [1] Howell, S. B. 2000, *Handbook of CCD Astronomy* (Cambridge, UK:
+        Cambridge University Press)
+    """
+    t = 300 * u.s
+    snr = 342 * np.sqrt(1 * u.electron)
+    gain = 5 * u.electron / u.adu
```
pllim:
```suggestion
    gain = 5 * (u.electron / u.adu)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    countrate = 24013 * u.adu * gain / t
+    npix = 1 * u.pixel
+    n_background = 200 * u.pixel
+    background_rate = 620 * u.adu / u.pixel * gain / t
```
pllim:
```suggestion
    background_rate = 620 * (u.adu / u.pixel) * gain / t
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    darkcurrent_rate = 22 * u.electron / u.pixel / u.hr
```
pllim:
```suggestion
    darkcurrent_rate = 22 * (u.electron / u.pixel / u.hr)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    readnoise = 5 * u.electron / u.pixel
```
pllim:
```suggestion
    readnoise = 5 * (u.electron / u.pixel)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    assert_quantity_allclose(result, t, atol=1 * u.s)
+
+
+def test_t_exp_analytic():
+    """
+    A test to check that the analytic method in
+    observation.exptime_from_howell_snr is done correctly.
+    """
+    snr_set = 50 * np.sqrt(1 * u.electron)
+    countrate = 1000 * u.electron / u.s
```
pllim:
```suggestion
    countrate = 1000 * (u.electron / u.s)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    npix = 1 * u.pixel
+    background_rate = 100 * u.electron / u.pixel / u.s
```
pllim:
```suggestion
    background_rate = 100 * (u.electron / u.pixel / u.s)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    darkcurrent_rate = 5 * u.electron / u.pixel / u.s
```
pllim:
```suggestion
    darkcurrent_rate = 5 * (u.electron / u.pixel / u.s)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +442,104 @@ def test_countrate_neg_leak():
...
+    readnoise = 1 * u.electron / u.pixel
```
pllim:
```suggestion
    readnoise = 1 * (u.electron / u.pixel)
```

synphot/observation.py
```diff
@@ -694,3 +694,220 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+        sqrt(electrons).
+    """
+    readnoise = _get_shotnoise(readnoise)
+    gain_err = _get_shotnoise(gain * ad_err)
+
+    pixel_terms = npix * (1 + npix / n_background)
+    detector_noise = (background + darkcurrent +
+                      readnoise ** 2 + gain_err ** 2)
+
+    sn = counts / np.sqrt(counts + pixel_terms * detector_noise)
```
eteq: (minor)

```suggestion
    return counts / np.sqrt(counts + pixel_terms * detector_noise)
```

synphot/observation.py
```diff
@@ -694,3 +694,220 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+
+    return sn
```
eteq: (goes with the above suggestion)
```suggestion
```

synphot/observation.py
```diff
@@ -694,3 +694,220 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+               gain=1 * (u.electron / u.adu),
+               ad_err=AD_ERR_DEFAULT):
+    """
+    A function to calculate the idealized theoretical signal to noise ratio
+    (SNR) of an astronomical observation with a given number of counts. This
+    expression is taken from pg 55 of [1]_).
+
+    Parameters
+    ----------
+    counts : `~astropy.units.Quantity`
```
eteq: This *might* be a matter of me missing a "higher" layer, but I think the more useful input here is *flux*.  That is, most astronomers don't know the number of counts, but they do know the flux of their target (or in the given spectral pixel, or whatever).

Or is the intent to have this be a "lower layer" that is called by other functions that just translate flux to counts?

synphot/observation.py
```diff
@@ -694,3 +694,220 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+    B = (-1) * snr ** 2 * (countrate + npix * (background_rate +
+                                               darkcurrent_rate))
+    C = (-1) * snr ** 2 * npix * readnoise ** 2
+
+    t = (-B + np.sqrt(B ** 2 - 4 * A * C)) / (2 * A)
+
+    if gain_err.value > 1 or not np.isfinite(n_background.value):
+        from scipy.optimize import fsolve
+        # solve t numerically
+        t = fsolve(_t_with_small_errs, t, args=(background_rate,
```
eteq: I'm confused by why `howell_snr` is not called in this function (and its use in `exptime_from_howell_snr`)? It would definitely be better to have a shared implementation between the two...

synphot/observation.py
tcjansen: Ah yeah, so the reason I do this is because of the "Returns" section in the docstring:
```
Returns
    -------
    sn : `~astropy.units.Quantity`
        The signal to noise ratio of the given observation in untis of
        sqrt(electrons).
```
I'm all for getting rid of the `sn` line. What's the right way of changing the docstring accordingly? (I could just get rid of "Returns" but I personally appreciate seeing this section in docstrings..)

synphot/observation.py
tcjansen: I might be misunderstanding the question but I think we use counts here because we only care about the number of photons hitting the ccd, not the total energy absorbed by the ccd (which is what I think of when I think of "flux", i.e. W/m^2). I do see how counts can be labeled as flux if you think of flux as just the "net amount of stuff going through an area", but I would easily confuse the parameter to mean "power density" if we were to change it.

synphot/observation.py
tcjansen: Hm, maybe the `from_howell_snr` bit in `exptime_from_howell_snr` is misleading, because the two snr expressions are slightly different - `howell_snr` assumes you already know the exposure time (i.e. you input `counts` instead of `countrate`, `darkcurrent` instead of `darkcurrent_rate`, etc), where the expression for the SNR in `exptime_from_howell_snr` is solving for `t` assuming you only know the count rates of each component.

Let me know if that doesn't answer the question and I'll give it another shot!

tcjansen: I guess I named it `exptime_from_howell_snr` instead of `exptime_from_snr` to be explicit that the expression comes from the CCD Handbook (and not just something I made up, hah), though I do see how the name leads one to think that this function uses the `howell_snr` function to get `t`. I'm open to suggestions for a different name!

tcjansen:
So upon including the signal to noise calculations in my previous synphot tutorials, I found that (as it stands now) the counts returned by synphot's `countrate` method are in incompatible units with `howell_snr` - that is, `howell_snr` requires the units be convertible to electrons, which `astropy.units.ct` are not, I think due to this comment regarding the output of `countrate`:

>The unit count may refer to different physical units for different instruments. For instance, in HST, it refers to the actual detector counts for the FOC, FOS, HRS, and HSP instruments. While for the WF/PC-1, WFPC2, NICMOS, WFC3, COS, ACS, and STIS instruments, it refers to electrons.

In other words (if I understand correctly) synphot counts can either mean ADU (as we saw in section 8 of [the first example](https://github.com/tcjansen/GSoC/blob/the-first-notebook/observing_examples.ipynb)) or electrons.

So what I could do is change the required unit convertability in the `u.quantity_input` decorator to `counts=u.ct`, but that leaves some uncomfortable ambiguity for the user because it requires them to be sure that their counts actually mean "electrons" before passing them into `howell_snr`. Hm...

synphot/observation.py
pllim: You can keep the "Returns" section as is. I do that sometimes too, naming a return variable but not really using it in the function.

synphot/observation.py
pllim: I guess a higher layer can be a user-facing example that sets up the Observation first, then calculates the count rate, and then pass the count rate into this function?

synphot/observation.py
pllim: https://twitter.com/codinghorror/status/506010907021828096?lang=en -- Erik, you got some name suggestions?

pllim:
Re: https://github.com/spacetelescope/synphot_refactor/pull/204#issuecomment-511968052

Yes, count rate can be ambiguous and really depends on how instrument scientists shipped their filter curves. So, in your workflow, adding an extra step to make sure count rate converts to electron definition (depending on the detector specs) before passing it into your function sounds reasonable. You can even add that as a note in your docstring.

synphot/observation.py
eteq: > I do see how counts can be labeled as flux if you think of flux as just the

Ah sorry, not quite what I meant here - rather I was thinking in the direction @pllim is saying, and I skipped a few steps.  That is, I agree we *don't* want to do the conversion you're saying here, @tcjansen, but rather what @pllim suggested, that this could be the "last" function in a set of calls.

In the end, this might end up being one of the "high-level" functions @tcjansen might adapt from some of the notebooks, I think. At some point it might be desirable to note that in the docs here, but I guess it's not good to do that until the higher level function exists to point the reader to.

synphot/observation.py
eteq: > the two SNR expressions are slightly different 

Ah, I see your point here.  It *feels* like it should be possible to unify these a bit, but it's not obvious to me how and the perfect is the enemy of the good!

eteq:
I was about to leave a note suggesting something like what @pllim said in https://github.com/spacetelescope/synphot_refactor/pull/204#issuecomment-511975595 before reading that comment, so I think that's a good idea!

I guess the key point for *this* PR is that the docstring should say something like "if you pass in something where that's not convertible to counts it won't work.  Be sure whatever you're using has already been turned into something that obeys poisson statistics", since I think that's the *real* constraint here.


One more related thought: instead of calling it `howell_*` perhaps it should be `poisson_*`?  That is, it's not really the fact that Howell wrote it down that's relevant to the science context, but rather the assumption that it's working on things that obey counting statistics.  (another option would be `ccd_*`, although I think that might be more specific than necessary).

tcjansen:
>  instead of calling it `howell_*` perhaps it should be `poisson_*`? That is, it's not really the fact that Howell wrote it down that's relevant to the science context, but rather the assumption that it's working on things that obey counting statistics. (another option would be `ccd_*`, although I think that might be more specific than necessary).

I think the specificity of ccd_* as a name is a good thing! That is, it's generally known that ccds are associated with poisson noise, where poisson noise can be associated with many other things. Thanks for the idea, I prefer ccd_* to howell_*

eteq:
From some out-of-band conversation with @tcjansen, I'm realizing maybe we should slightly loosen what I said in https://github.com/spacetelescope/synphot_refactor/pull/204#issuecomment-514324909, though.  Perhaps we should allow `counts`, `photon`, `electron` as valid units, as long as they are consistent?  I.e., it's not OK to mix counts and photons, but just photons or just electrons (and the appropriate rates) is ok?  The easy way to do that might be do leave things as they stand (using counts) but just have a converter along  the lines of:
```
if q.unit in (u.electron, u.photon):
    return u.Quantity(q.value, u.count)
elif q.unit in (u.electron/u.s, u.photon/u.s):
    return u.Quantity(q.value, u.count/u.s)
else:
    raise ... some error ...
```

synphot/observation.py
```diff
@@ -694,3 +694,245 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+
+    # solve t with the quadratic equation (pg. 57 of Howell 2000)
+    A = countrate ** 2
+    B = (-1) * snr ** 2 * (countrate + npix * (background_rate +
+                                               darkcurrent_rate))
+    C = (-1) * snr ** 2 * npix * readnoise ** 2
+
+    t = (-B + np.sqrt(B ** 2 - 4 * A * C)) / (2 * A)
+
+    if gain_err.value > 1 or np.isfinite(n_background.value):
```
bmorris3: I thin this is one of those places where `gain_err.value` should be `gain_err.to(<some units>).value > 1`.

bmorris3: Same for `n_background.value`.

synphot/observation.py
```diff
@@ -694,3 +694,245 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+        (1995)
+
+    Returns
+    -------
+    t : `~astropy.units.Quantity`
+        The exposure time needed (in seconds) to achieve the given signal
+        to noise ratio.
+    """
+    if countrate.unit in (u.electron / u.s, u.photon / u.s):
+        countrate = countrate.value * u.ct / u.s
```
bmorris3: Same comment as earlier: https://github.com/spacetelescope/synphot_refactor/pull/204/files#r306842663

bmorris3:
Looks good to me! 👍

synphot/tests/test_observation.py
```diff
@@ -439,3 +440,136 @@ def test_countrate_neg_leak():
...
+    assert_quantity_allclose(snr_calc, snr_set,
+                             atol=0.5 * u.Unit(""))
+
+
+def test_snr_with_countrate():
+    """
+    A test to make sure `ccd_snr` works with the units that
+    `countrate` returns.
+    """
+    source_spec = SourceSpectrum(BlackBodyNorm1D)
```
pllim: For sanity check, we can set this up so that the observation would produce appox. 1 counts/s.
```suggestion
    area = 1 * u.m ** 2
    bp = SpectralElement(Const1D)
    source_spec = SourceSpectrum(BlackBodyNorm1D)
    source_spec = source_spec.normalize(
        1 * u.ct, bp, force='extrap', wavelengths=source_spec.waveset, area=area)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +440,136 @@ def test_countrate_neg_leak():
...
+    bp = SpectralElement(Const1D)
```
pllim:
```suggestion
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +440,136 @@ def test_countrate_neg_leak():
...
+    obs = Observation(source_spec, bp, force='extrap')
+    counts = obs.countrate(1 * u.m ** 2) * 1 * u.s
```
pllim:
```suggestion
    counts = obs.countrate(area) * u.s
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +440,136 @@ def test_countrate_neg_leak():
...
+    snr = ccd_snr(counts)
+
+    assert snr.unit == u.Unit("")
```
pllim: I think SNR should be close to 1, as `1/sqrt(1)` is 1, but I'll leave that as an exercise for you.
```suggestion
    assert snr.unit == u.Unit("")
    np.testing.assert_allclose(snr.value, ???)
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +440,136 @@ def test_countrate_neg_leak():
...
+
+
+@raises(u.UnitsError)
+def test_counts_units():
+    """
+    A test to make sure `ccd_snr` raises a UnitError if the user
```
pllim:
```suggestion
    A test to make sure ``ccd_snr`` raises a UnitError if the user
```

synphot/tests/test_observation.py
```diff
@@ -439,3 +440,136 @@ def test_countrate_neg_leak():
...
+    gives an incompatible unit for counts.
+    """
+    return ccd_snr(1 * u.m)
+
+
+@raises(u.UnitsError)
+def test_countrate_units():
+    """
+    A test to make sure `exptime_from_ccd_snr` raises a UnitError
```
pllim:
```suggestion
    A test to make sure ``exptime_from_ccd_snr`` raises a UnitError
```

synphot/observation.py
tcjansen: > I thin this is one of those places where `gain_err.value` should be `gain_err.to(<some units>).value > 1`.

Oh yeah, this is taken care of in the call to `_get_shotnoise` in line 892.

> Same for `n_background.value`.

Does `n_background` have to be in a certain unit if we're just checking whether or not it's a finite number?

synphot/observation.py
```diff
@@ -694,3 +694,245 @@ def as_spectrum(self, binned=True, wavelengths=None):
...
+    t = (-B + np.sqrt(B ** 2 - 4 * A * C)) / (2 * A)
+
+    if gain_err.value > 1 or np.isfinite(n_background.value):
+        from scipy.optimize import fsolve
+        # solve t numerically
+        t = fsolve(_t_with_small_errs, t, args=(background_rate,
+                                                darkcurrent_rate,
+                                                gain_err, readnoise, countrate,
+                                                npix, n_background))
+        t = float(t) * u.s
```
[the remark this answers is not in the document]
tcjansen: Ah, because `fsolve` returns an array. So to match the return type when `fsolve` isn't called, I convert it into a float. Though this does bring up a good point, that is, since `fsolve` returns the roots of the function (the function in this case being a quadratic), will `fsolve` ever return two solutions...? I'll have to think about this for a minute

synphot/tests/test_observation.py
tcjansen: Ah I'm glad you suggested this because it didn't pass! And it's because I set the default gain to 1 ct/adu and the default ad_err to some other number, so for very small counts (like 1 in this test) the gain error significantly changes the snr, even if it isn't specified by the user. So to keep up with the "idealized" snr that we represent by default, I changed the default gain to 0 such that its error is ignored unless specified.

pllim: Well, if 1 count is an edge case that requires gain of zero, you could also consider normalizing to higher count values. What do you think?

tcjansen: Nah, I think it's good to test even these edge cases. Including the gain error by default doesn't make too much sense anyway since in most cases it will probably be ignored.

synphot/observation.py
pllim: Has this been answered, @tcjansen ?

tcjansen: No, I've left it as is

tcjansen:
All of the changes made in this PR have been moved to #207

PR #207, opened 2019-07-30
WIP - A higher order function for wrapping synphot functionality

tcjansen:
The initial state of this PR is beyond a work in progress if you know what I mean 😅, but I wanted to put it out there so we can chat about it in our GSoC meeting.

So I'm envisioning some kind of wrapper for synphot in which you can initiate a class with the source spectrum of an object, give a telescope diameter, and boom, you can get attributes like SNR, countrate, exptime, etc...:
```
>>> from synphot.spectrum import SourceSpectrum
>>> from synphot.wrapper import Observe
>>> import astropy.units as u

>>> tele_diameter = 1 * u.m
>>> source_spec = SourceSpectrum(...)
>>> some_observation = Observe(source_spec, tele_diameter)

>>> some_observation.countrate
1e6 ct/s
>>> some_observation.snr
100
...
``` 
The observation can be customized by setting a bandpass, giving a desired SNR (to get an exposure time), providing the quantum efficiency of the CCD or an atmospheric transmission model, etc.

### DIFF
--- a/synphot/experimental/exposure.py
+++ b/synphot/experimental/exposure.py
@@ -1,0 +1,269 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""This module is a wrapper for synphot's most commonly used functions.
+
+"""
+
+# THIRD-PARTY
+import numpy as np
+
+# ASTROPY
+from astropy import units as u
+from astropy.modeling.models import Const1D
+
+# LOCAL
+from synphot.models import Empirical1D
+from synphot.spectrum import SpectralElement
+from .observation import Observation, ccd_snr, exptime_from_ccd_snr
+from synphot.experimental.skymodel import atmospheric_transmittance
+from synphot import exceptions
+
+__all__ = ['Exposure']
+
+
+AD_ERR_DEFAULT = np.sqrt(0.289) * (u.adu / u.pixel)
+
+
+class Exposure(object):
+    """
+    Parameters
+    ----------
+    source_spec : `~synphot.spectrum.SourceSpectrum`
+        The spectrum of the target object scaled by distance (i.e. the
+        target spectrum just before interacting with the instrument).
+    diameter : `~astropy.units.quantity.Quantity`
+        Diameter of the telescope aperture.
+    bandpass : str, list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values, or
+        `~synphot.spectrum.SpectralElement`
+        The bandpass(es) of this observation. Can either be specified by a
+        string or a list with dimensions 2 x n, where one row is the
+        list of wavelengths associated with the bandpass, and the other
+        row contains the fractional values of the bandpass.
+        Default is a constant model with amplitude = 1, i.e. as if a
+        bandpass with perfect throughput were in effect.
+    quantum_efficiency : bool or list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values
+        A model for the quantum efficiency of the instrument's CCD.
+        Default is False. If set to True, quantum_efficiency assumes
+        a generic silicon CCD response function.
+    atmospheric_transmission : bool or list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values
+        A model for the atmopsheric transmission. Default is False.
+        If set to True, atmospheric_transmission queries from SkyCalc
+        with default parameters. For more information on SkyCalc, see:
+        (http://www.eso.org/observing/etc/bin/gen/form?INS.MODE=
+        swspectr+INS.NAME=SKYCALC)
+        For more information on the default parameters, see:
+        (https://www.eso.org/observing/etc/doc/skycalc/helpskycalccli.
+        html#skycalc-input-parameters)
+    force : {None, ‘none’, ‘extrap’, ‘taper’}
+        Force the creation of an observation even when source spectrum
+        and bandpass do not fully overlap:
+
+            * `None` or 'none' - Source must encompass bandpass (default)
+            * 'extrap' - Extrapolate source spectrum (this changes the
+              underlying model of ``spec`` to always extrapolate,
+              if applicable)
+            * 'taper' - Taper source spectrum to zero
+    """
+    @u.quantity_input(diameter=u.m, exptime=u.s)
+    def __init__(self, source_spec, diameter, exptime,
+                 bandpass=SpectralElement(Const1D, amplitude=1),
+                 quantum_efficiency=False, atmospheric_transmission=False,
+                 force=None):
+        self.source_spec = source_spec
+        self.diameter = diameter
+        self.exptime = exptime
+        self._bandpass = bandpass
+        self._qe = quantum_efficiency
+        self._atmo = atmospheric_transmission
+
+        # turn spectrum elements into synphot objects
+        self._set_bandpass()
+
+        if type(self._qe) == list:
+            self._custom_qe()
+        elif self._qe is True:
+            self._generic_ccd()
+
+        if type(self._atmo) == list:
+            self._custom_atmosphere()
+        elif self._atmo is True:
+            self.atmospheric_transmission = atmospheric_transmittance()
+
+        self.observation = Observation(self.source_spec, self.throughput,
+                                       force=force)
+
+    @property
+    def aperture_area(self):
+        """ Area of the telescope aperture. """
+        return np.pi * (self.diameter / 2) ** 2
+
+    @property
+    def countrate(self):
+        """ Countrate of the observation in ct / s."""
+        return self.observation.countrate(area=self.aperture_area)
+
+    def _set_lists(self, _list, _list_str):
+        """
+        Turns user-provided models into SpectralElements.
+        Written such that the order in which the waveset and corresponding
+        values are given in the list does not matter.
+        """
+        if len(_list) != 2:
+            raise exceptions.BandpassError(_list_str + " should be "
+                                           "a list with 2 x n "
+                                           "dimensions.")
+        try:
+            __list = _call_SpecElement_4list(_list[0], _list[1])
+        except u.UnitsError:
+            try:
+                __list = _call_SpecElement_4list(_list[1], _list[0])
+            except u.UnitsError:
+                print("rows of " + _list_str + " list must be of type "
+                      "astropy.units.quantity.Quantity")
+        return __list
+
+    def _set_bandpass(self):
+        """
+        Sets the SpectralElement for the bandpass, which can be one of the
+        given synphot bandpasses, a custom user-provided model, or one already
+        set as a SpectralElement.
+        """
+        if type(self._bandpass) == list:
+            self.bandpass = self._set_lists(self._bandpass, 'bandpass')
+
+        elif type(self._bandpass) == str:
+            self.bandpass = SpectralElement.from_filter(self._bandpass)
+
+        elif type(self._bandpass) == SpectralElement:
+            self.bandpass = self._bandpass
+
+        else:
+            raise exceptions.BandpassError("bandpass must be of type str, "
+                                           "`~synphot.spectrum."
+                                           "SpectralElement`, or list of "
+                                           "shape(2, n) containing `~astropy"
+                                           ".units.quantity.Quantity` values"
+                                           )
+
+        self.throughput = self.bandpass
+
+    def _custom_qe(self):
+        """
+        Sets the SpectralElement for a given quantum efficiency and convolves
+        it with the total throughput.
+        """
+        self.quantum_efficiency = self._set_lists(self._qe,
+                                                  'quantum_efficiency')
+        self.throughput *= self.quantum_efficiency
+
+    def _generic_ccd(self):
+        """
+        Sets the SpectralElement for a generic throughput function of
+        a silicon detector and convolves it with the total throughput.
+        """
+        wl = [100., 309.45600215, 319.53901519, 329.99173871,
+              339.70504127, 349.78805431, 379.53294279, 390.00376402,
+              399.57293121, 409.3114413, 419.5961146, 429.14136695,
+              439.52687038, 449.9795939, 459.69289647, 469.60785929,
+              479.85892255, 489.5638226, 499.59835962, 509.99161922,
+              519.5607864, 529.30446727, 539.85285015, 549.59976275,
+              559.51472558, 569.94676599, 579.68075166, 589.59571448,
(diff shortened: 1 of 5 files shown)


synphot/tests/test_wrapper.py
```diff
@@ -0,0 +1,48 @@
+from synphot.models import BlackBody1D
+from synphot.spectrum import SourceSpectrum, SpectralElement
+from synphot.wrapper import Observe
+import astropy.units as u
+import numpy as np
+
+
+class TestObserve:
```
bmorris3: It's not strictly necessary, but one style point: it's more explicit to write `class TestObserve(object)`, implying that `TestObserve` inherits from `object`.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+from astropy.modeling.models import Const1D
+
+# LOCAL
+from . import binning, exceptions, units, utils
+from .models import Empirical1D
+from .spectrum import BaseSourceSpectrum, SourceSpectrum, SpectralElement
+from .observation import Observation, ccd_snr, exptime_from_ccd_snr
+
+
+class Observe:
```
bmorris3: Same comment as earlier, prefer `class Observe(object):` over `class Observer`. 

***

Sidebar: perhaps this should be `Expose` rather than `Observe` since synphot already has an `Observation` class.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        Default is None.
+    atmospheric_transmission : list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values
+        A model for the atmopsheric transmission.
+    """
+    @u.quantity_input(diameter=u.m, distance=u.pc, exptime=u.s)
+    def __init__(self, source_spec, diameter, distance=0 * u.pc,
+                 bandpass=SpectralElement(Const1D, amplitude=1),
+                 quantum_efficiency=None, atmospheric_transmission=None,
+                 force=None, exptime=1 * u.s):
```
bmorris3: Should things like exptime have a default values? I think that's a thing that will really need to be supplied in every case, right? And distance sticks out as bing an odd one too, nothing you observe is actually at 0 pc.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        self.source_spec = source_spec
+        self.diameter = diameter
+        self.aperture_area = np.pi * (diameter / 2) ** 2
+        self.distance = distance
+        self._bandpass = bandpass
+        self._qe = quantum_efficiency
+        self._at = atmospheric_transmission
```
bmorris3: Just for clarity, I vote for `self._atmo`, since `at` is also an English word.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        A model for the quantum efficiency of the instrument's CCD. 
+        Default is None.
+    atmospheric_transmission : list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values
+        A model for the atmopsheric transmission.
+    """
+    @u.quantity_input(diameter=u.m, distance=u.pc, exptime=u.s)
+    def __init__(self, source_spec, diameter, distance=0 * u.pc,
+                 bandpass=SpectralElement(Const1D, amplitude=1),
+                 quantum_efficiency=None, atmospheric_transmission=None,
```
bmorris3: I think this might be controversial (👀 @eteq), but I might vote for atmospheric transmission to take a _boolean_ input, and if `True`, return the skycalc transmittance curve with some default parameters; if `False`, don't include atmospheric transmission. I think for 90% of users that will give them the answer they want within a factor of 2 with as little data wrangling as possible.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        # turn spectrum elements into synphot objects
+        self._set_bandpass()
+        if self._qe is not None:
+            self._include_qe()
+        if self._at is not None:
+            self._include_atmosphere()
+                    
+        self.obs = Observation(self.source_spec, self.throughput, force=force)
+        self.countrate = self.obs.countrate(area=self.aperture_area)
+        self.snr = ccd_snr(self.countrate * self.exptime)
```
bmorris3: I think since `ccd_snr` takes lots of optional arguments, you could instead make SNR into a method, i.e.: `def snr(self, <args>, <kwargs>)`.

synphot/wrapper.py
bmorris3: I might even vote the same thing for `quantum_efficiency`. Perhaps it could instead be `silicon_detector=True`, and if `True`, return the generic CCD response curve that we have on hand. If `False`, assume a constant throughput.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+                "`~astropy.units.quantity.Quantity` values")
+
+        self.throughput = self.bandpass
+
+    def _include_qe(self):
+        if len(self._qe) != 2:
+            raise exceptions.BandpassError("quantum_efficiency should be "
+                                           "a list with 2 x n "
+                                           "dimensions.")
+        try:
```
bmorris3: Reminder RE: slack convo to let the function fail when the user supplies the wrong inputs, rather than writing the same exception block in a bunch of places. Bumpers are nice but at some point obfuscate your code.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        except u.UnitsError:
+            try:
+                self.atmospheric_transmission = _call_SpecElement_4list(self._at[1], self._at[0])
+            except u.UnitsError:
+                print("rows of atmospheric_transmission list must be of type "
+                      "astropy.units.quantity.Quantity")
+
+        self.throughput *= self.atmospheric_transmission
+
+@u.quantity_input(waves=u.angstrom, frac=u.Unit(''))
```
bmorris3: I don't think a private method needs a decorator to enforce that you're using it correctly. Ideally the user will never know this function exists, so they shouldn't need forceful reminders to get their units correct).

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        if self._qe is not None:
+            self._include_qe()
+        if self._at is not None:
+            self._include_atmosphere()
+                    
+        self.obs = Observation(self.source_spec, self.throughput, force=force)
+        self.countrate = self.obs.countrate(area=self.aperture_area)
+        self.snr = ccd_snr(self.countrate * self.exptime)
+
+    def _set_bandpass(self):
```
bmorris3: You don't strictly need to write docstrings for private functions, and you certainly don't need to write them with the same rigor as you do for public functions, but it's still nice to have a one-line summary for each method, even the private ones. This happens all over the place in astropy, [for example](https://github.com/astropy/astropy/blob/master/astropy/coordinates/sky_coordinate.py#L302-L326).

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        elif type(self._bandpass) == SpectralElement:
+            self.bandpass = self._bandpass
+
+        else:
+            raise exceptions.BandpassError("bandpass must be of type str, "
+                "`~synphot.spectrum.SpectralElement`, or "
+                "list of shape(2, n) containing "
+                "`~astropy.units.quantity.Quantity` values")
+
+        self.throughput = self.bandpass
```
bmorris3: I might be missing something here, but is there a point of having a separate `throughput` and `bandpass` object carried around if one becomes the other eventually?

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        Default is a constant model with amplitude = 1, i.e. as if there
+        were no bandpass in effect.
+    quantum_efficiency : list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values
+        A model for the quantum efficiency of the instrument's CCD. 
+        Default is None.
+    atmospheric_transmission : list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values
+        A model for the atmopsheric transmission.
+    """
```
bmorris3: Don't forget to document what `force` does in this docstring – that's always been a bit of black magic to me.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # turn spectrum elements into synphot objects
+        self._set_bandpass()
+        if self._qe is not None:
+            self._include_qe()
+        if self._at is not None:
+            self._include_atmosphere()
+                    
+        self.obs = Observation(self.source_spec, self.throughput, force=force)
+        self.countrate = self.obs.countrate(area=self.aperture_area)
```
bmorris3: RE slack discussion: you could choose to make `countrate` a property which only gets computed when the user calls it explicitly, using the `@property` decorator.

synphot/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        Diameter of the telescope aperture.
+    bandpass : str, list with dimension 2 x n containing
+        `~astropy.units.quantity.Quantity` values, or 
+        `~synphot.spectrum.SpectralElement`
+        The bandpass(es) of this observation. Can either be specified by a 
+        string or a list with dimensions 2 x n, where one row is the
+        list of wavelengths associated with the bandpass, and the other
+        row contains the fractional values of the bandpass.
+        Default is a constant model with amplitude = 1, i.e. as if there
+        were no bandpass in effect.
```
bmorris3: Rather than calling this `no bandpass`, perhaps say something like `perfect throughput`?

synphot/wrapper.py
tcjansen: Okay, I agree with this. Though I am curious what the controversy is..!

pllim:
Thanks! I don't have a chance to do a detailed review yet, but I would propose renaming `wrapper.py` to something less cryptic. How about `groundbased` or something to explain at a high level the intent of your new module?

synphot/wrapper.py
tcjansen: Sure, I agree with `exptime` not being a default param.

As for `distance`, I set the default to 0 instead of something like `False` for a couple of reasons... first as a way to indicate that the user should have already scaled the source spectrum for distance (i.e. `source_spec` should be as it is just before it hits the instrument, i.e. 0 pc), and also because while scaling the spectrum for a star is straight forward ([R / d]^2), I wasn't sure how exactly to do that for something like a galaxy so I put it off for the moment 😂

tcjansen: I guess I was envisioning that the user could give the source spectrum as it is at the object, and `Expose` could scale the spectrum given a distance. But on second thought, I think it would be better for all parties if we get rid of `distance` and just specify that `Expose` assumes the user has already scaled the spectrum.

synphot/wrapper.py
tcjansen: Since I've moved these bits to a function (where the exception is only stated once) it's okay to keep?

synphot/wrapper.py
tcjansen: Ah yeah, I put that decorator where it is because there doesn't seem to be a better place to put it. Passing these parameters in as lists is not the default option, so I couldn't put it before `__init__()`, but I do still want to enforce that the user's custom models are in the correct units.

synphot/wrapper.py
tcjansen: This comes from me wanting the user to be able to call `self.bandpass` to inspect the filter, while `self.throughput` would allow them to see all of the throughput effects in play (QE, atmosphere, etc).

tcjansen:
Does `expose.py` work? It follows some other `synphot` style choices as well (e.g. `synphot.observe.Observe`)

synphot/expose.py
```diff
@@ -0,0 +1,439 @@
...
+        params['observatory'] = '2640'
+    elif (params['observatory'] == '3060m' or
+          params['observatory'] == 'armazones'):
+        params['observatory'] = '3060'
+    else:
+        raise ValueError('Wrong Observatory name, please refer to the '
+                         'skycalc_cli documentation.')
+
+    # Use the bit from skycalc_cli which queries from the SkyCalc Sky Model
+    server = 'http://etimecalret-001.eso.org'
```
pllim: Why is this not an `astroquery` call?

CHANGES.rst
```diff
@@ -3,6 +3,11 @@
 
 - Removed Python 2 support. This version is only compatible with Python 3.5
   or later. [#185]
+- New ``synphot.observation.ccd_snr()`` function to calculate the signal
+  to noise ratio given a count rate and exposure time. [#204]
```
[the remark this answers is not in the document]
pllim: Ignore this. I didn't realize you had commits from #204 when I wrote this. And I thought I removed it before posting but obviously I didn't. :grimacing:

CHANGES.rst
```diff
@@ -3,6 +3,11 @@
...
+- New ``synphot.observation.exptime_from_ccd_snr()`` function to
+  determine the exposure time needed to achieve a given signal to noise
+  ratio. [#204]
```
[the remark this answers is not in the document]
pllim: Ignore this. I didn't realize you had commits from #204 when I wrote this. And I thought I removed it before posting but obviously I didn't. :grimacing:

tcjansen:
So I included changes from #204 because this PR requires those changes in order to run properly (to call `ccd_snr` for example). Is it not a problem to leave those changes out?

tcjansen:
As for the name changes, maybe `exposure.py` and `Exposure`? `groundbased.py` would be misleading because it doesn't actually require atmospheric transmission.

pllim:
@tcjansen 

* How about we make a new "experimental" folder under "synphot" and then you can name it `exposure.py` in that folder? In the future, when the API matures, I can rethink on where to house it permanently. 

* Are you suggesting that we merged #204 into this PR? If so, please close #204 without merge.

Thanks!

synphot/expose.py
tcjansen: That bit is an artifact from me borrowing straight from skycalc_cli. Is there an astropy equivalent to `requests`?

pllim: There is a `astropy.utils.data.download_file()` function but I think @bmorris3 said that it is not quite what you need. So my other suggestion would be to move the skycalc query out to its own module (after you created this "experimental" folder and move your work there).

pllim:
As per the telecon today, @tcjansen will move these new functionalities to a new subpackage in `astroplan` instead, with `synphot` being an *optional* dependency over there. cc @bmorris3 and @eteq 

Thank you for your continued patience, Tiffany!

pllim:
Since the astroplan PR is live, I am closing this. Thanks again, Tiffany!

PR #211, opened 2019-08-16
add option for download_data to use the astropy cache

### LINKED ISSUE #188: download_data does not download anything - skips everything

### LINKED ISSUE #208: have an option in download_data to download everything to the cache
Currently you have to specify a path for `download_data`.  That's fine for the use case illustrated in the docs, but if you don't use `download_data` but use some of the data files, the end result is a bunch of stuff in the cache. It would be nice if there was a way to instead tell `download_data` "download everything you need to the cache, because I'm not ready to work with it yet but am about to get on a flight (or otherwise be offline)".

I'll try to take a shot at this but putting it in now in case I don't find the time for a while.

eteq:
This PR closes #208 by making it possible to use the astropy download cache.  

One caveat: There are no unit tests.  I started trying to make one but then realized it's not clear to me that we can really test this properly without accidentally messing with the users' "true" cache.  The related tests in astropy test the cache machinery by downloading a "fake" file and then clearing just that file, but here we can't do it because we actually want the *correct* file.  So that's awkward. I can concoct some sort of complicated scheme like what we *used* to have in astropy where we fake the cache location, run the test, and then un-fake it, but that seems like it might be overkill for this.  @pllim do you agree and/or have a simpler idea for how I might test this?  Or alternatively I can add a quick test that would have the side-effect of potentially deleting the user's *actual* copy of the data (which is normally considered bad behavior for a test...)

(Note I did test it in the sense of actually running it myself... )

EDIT: Also fix #188

### DIFF
--- a/synphot/utils.py
+++ b/synphot/utils.py
@@ -252,14 +252,20 @@ def merge_wavelengths(waveset1, waveset2, threshold=1e-12):
     return out_wavelengths
 
 
-def download_data(cdbs_root, verbose=True, dry_run=False):
-    """Download CDBS data files to given root directory.
+def download_data(path_root=None, verbose=True, dry_run=False):
+    """Download synphot data files to given root directory or the astropy cache.
     Download is skipped if a data file already exists.
+
+    Note that if you have changed the path for the data files by providing a
+    synphot.cfg file pointing to local file, this function will not work. You
+    will need to have the configuration items point to wherever you wish to
+    download them from for this to work.
 
     Parameters
     ----------
-    cdbs_root : str
-        Root directory for CDBS data files.
+    path_root : str or None, optional
+        Root directory for data files. If None, download to the astropy cache
+        location instead of a specific directory.
 
     verbose : bool
         Print extra information to screen.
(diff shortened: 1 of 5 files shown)


pllim:
Thanks! The failures are related to upstream refactoring, tsk tsk. I'll have to investigate that first before I come back to this. :grimacing:

pllim:
I'll push follow-up commits and wrap this up. I think testing is possible if we stick to dry run mode and maybe some Mock (not sure but will try).

synphot/utils.py
```diff
@@ -252,14 +252,20 @@ def merge_wavelengths(waveset1, waveset2, threshold=1e-12):
     return out_wavelengths
 
 
-def download_data(cdbs_root, verbose=True, dry_run=False):
-    """Download CDBS data files to given root directory.
+def download_data(path_root=None, verbose=True, dry_run=False):
```
pllim: I am reverting the code to make `path_root` optional. It changes the default behavior and might confuse users.

pllim:
I cannot find an easy way for "cache only" option to work with `synphot.cfg`, so I updated to the doc to tell user to make a choice.

synphot/utils.py
```diff
@@ -282,55 +290,65 @@ def download_data(cdbs_root, verbose=True, dry_run=False):
...
+    file_list = []
 
     # See https://github.com/astropy/astropy/issues/8524
     for cfgitem in conf.__class__.__dict__.values():
         if (not isinstance(cfgitem, ConfigItem) or
                 not cfgitem.name.endswith('file')):
             continue
 
-        url = cfgitem()
+        url = cfgitem.defaultvalue
```
pllim: This line fixes #188 but also changes the behavior reflected in the diff for `test_download_data`. I think the behavior change is a small price to pay for this fix.

PR #236, opened 2020-01-15
Fix MANIFEST syntax

pllim:
My bad...


pllim:
`conda` build for 0.2.1 turns out to be working fine despite this bug, so I am moving the milestone to 0.3. There is no need to do a release just for this.

PR #239, opened 2020-01-29
DOC: Add mention of stsynphot Appendix B

### LINKED ISSUE #186: Origin of filter response functions
I am exploring using synphot for doing non-HST synthetic photometry.  I see there are a few response functions included - e.g., Johnson B.  Is it documented somewhere the origin of this response function (and the others)?  I ask as this response function is different than what I've used for sometime that is based on Bessell et al. 1998 (https://ui.adsabs.harvard.edu/abs/1998A%26A...333..231B/abstract).

pllim:
Fix #186 . Does this make it clearer, @karllark ?

Note to self: Check rendered RTD build in this PR.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -126,7 +126,8 @@ man_pages = [('index', project.lower(), project + u' Documentation',
 # -- Options for intersphinx --------------------------------------------------
 
 intersphinx_mapping.update({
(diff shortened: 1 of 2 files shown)


karllark:
Yes, this addresses the issue.

PR #243, opened 2020-02-11
ENH: specutils bridge

### LINKED ISSUE #43: Consolidate with specutils
There might be some overlap with `specutils`, which is an affiliated `astropy` package. When this package is stable, consolidate redundancy between its API and ours.

http://specutils.readthedocs.org/en/latest/
https://github.com/astropy/specutils

pllim:
Fix #43 

**TODO**

- [x] Write tests
- [x] Write user-facing doc
- [x] Get approval from `specutils` people.
  - [x] Do we need to support bandpass?
  - [x] Did I miss any other `Spectrum1D` use cases?

### DIFF
--- a/docs/synphot/bandpass.rst
+++ b/docs/synphot/bandpass.rst
@@ -14,6 +14,7 @@ A bandpass can be constructed by one of the following methods:
   keywords needed to define it into a
   :class:`~synphot.spectrum.SpectralElement` object.
 * Build a composite bandpass using :ref:`synphot-spec-math-op`.
+* Build an empirical bandpass using :ref:`synphot-bandpass-specutils`.
 
 It has various :ref:`photometric properties <synphot_formulae>` and these
 main components:
@@ -269,3 +270,36 @@ set at 0.8:
     from synphot import SpectralElement
     bp = SpectralElement(Const1D, amplitude=0.8)
     bp.plot([1000, 10000], title='Flat Bandpass')
+
+
+.. _synphot-bandpass-specutils:
+
+specutils
+---------
+
+A bandpass can be constructed from and written to `specutils.Spectrum1D`
+object. See :ref:`specutils documentation <specutils:specutils>` for more
+information on how to use `~specutils.Spectrum1D`.
+
+The example below writes a :ref:`synphot-box-bandpass` to a
+`~specutils.Spectrum1D` object:
+
+.. doctest-requires:: specutils
+
+    >>> from astropy import units as u
+    >>> from synphot import SpectralElement
+    >>> from synphot.models import Box1D
+    >>> bp = SpectralElement(Box1D, amplitude=1, x_0=600*u.nm, width=10*u.nm)
+    >>> spec = bp.to_Spectrum1D()
+
+Meanwhile, this example reads in a bandpass from a
+`~specutils.Spectrum1D` object:
+
+.. doctest-requires:: specutils
+
+    >>> from astropy import units as u
+    >>> from specutils import Spectrum1D
+    >>> from synphot import SpectralElement
+    >>> spec = Spectrum1D(spectral_axis=[100, 300]*u.nm,
+    ...                   flux=[0.1, 0.8]*u.dimensionless_unscaled)
+    >>> bp = SpectralElement.from_Spectrum1D(spec)
--- a/docs/synphot/bandpass.rst
+++ b/docs/synphot/bandpass.rst
@@ -290,7 +290,7 @@ The example below writes a :ref:`synphot-box-bandpass` to a
     >>> from synphot import SpectralElement
     >>> from synphot.models import Box1D
     >>> bp = SpectralElement(Box1D, amplitude=1, x_0=600*u.nm, width=10*u.nm)
-    >>> spec = bp.to_Spectrum1D()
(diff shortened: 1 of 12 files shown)


pllim:
This is ready for review.

docs/synphot/bandpass.rst
```diff
@@ -269,3 +270,36 @@ set at 0.8:
...
+Meanwhile, this example reads in a bandpass from a
+`~specutils.Spectrum1D` object:
+
+.. doctest-requires:: specutils
+
+    >>> from astropy import units as u
+    >>> from specutils import Spectrum1D
+    >>> from synphot import SpectralElement
+    >>> spec = Spectrum1D(spectral_axis=[100, 300]*u.nm,
+    ...                   flux=[0.1, 0.8]*u.dimensionless_unscaled)
```
eteq: Question which may or may not need changing: what happens if I pass in a spectrum that is *not* `dimensionless_unscaled`?  I don't see anywhere that this is checked for whether or not the unit makes sense.  I'm not sure it needs to be, but it might be a good idea since otherwise the units will get all funky, right?

synphot/spectrum.py
```diff
@@ -778,6 +778,67 @@ def plot(self, wavelengths=None, **kwargs):  # pragma: no cover
...
+
+        Returns
+        -------
+        sp : `BaseSourceSpectrum`
+            Empirical spectrum.
+
+        """
+        # Spectrum1D is designed to be immutable, so no need to make
+        # copies of spectral_axis nor flux.
+        return cls(Empirical1D, points=spec.spectral_axis,
```
eteq: This doesn't seem to do anything with `mask`. Am I right that `Empirical1D` has no concept of a mask?  If so, it probably makes sense to apply the mask to the spectral points if they exist.  Something like:
```suggestion
        if spec.mask is not None:
            msk = spec.mask == False
            lookup_table = spec.flux[msk]
            points = spec.spectral_axis[msk]
        else:
            lookup_table = spec.flux
            points = spec.spectral_axis
        return cls(Empirical1D, points=points, lookup_table=lookup_table,
```

?

synphot/spectrum.py
```diff
@@ -778,6 +778,67 @@ def plot(self, wavelengths=None, **kwargs):  # pragma: no cover
...
+                   lookup_table=spec.flux, keep_neg=keep_neg,
+                   meta={'header': spec.meta.copy()})
+
+    def to_Spectrum1D(self, wavelengths=None, **kwargs):
```
eteq: I'm not sure I like the look of this where it has a weird mix of casing conventions.  What about:
```suggestion
    def to_spectrum1d(self, wavelengths=None, **kwargs):
```
?  Or even:
```suggestion
    def to_specutils(self, wavelengths=None, **kwargs):
```
?

synphot/spectrum.py
```diff
@@ -778,6 +778,67 @@ def plot(self, wavelengths=None, **kwargs):  # pragma: no cover
         w, y = self._get_arrays(wavelengths)
         self._do_plot(w, y, **kwargs)
 
+    @classmethod
+    def from_Spectrum1D(cls, spec, keep_neg=False):
```
eteq: To go along with my previous comment, this I guess would be 
```suggestion
    def from_spectrum1d(cls, spec, keep_neg=False):
```
or 
```suggestion
    def from_specutils(cls, spec, keep_neg=False):
```
?

docs/synphot/bandpass.rst
pllim: Unit checking is done inside `SpectralElement` and this is already tested somewhere else. You would get this error if you pass in something with dimension (e.g., Jy):
```
UnitConversionError: 'Jy' (spectral flux density) and '' (dimensionless) are not convertible
```

synphot/spectrum.py
pllim: Oh, I am not aware you have masking. I took the API off Getting Started and did not dig too deep into `specutils`. 😬  I'll add this in. Thanks!

synphot/spectrum.py
pllim: I debated with myself about this. The class is technically `Spectrum1D` with uppercase letters but that is also against PEP 8 naming convention for method names. Since you prefer all lowercase, I'll make that change.

However, I do not think it should be `to_specutils`. What if you have a different class in `specutils` that you wish to tie in to `synphot` in the future? Do you want this method to handle all those cases?

pllim: Renamed to lowercase.

synphot/spectrum.py
pllim: Wait a minute... Why isn't the mask automatically applied by `specutils` when you access the `flux` or `spectral_axis` attributes?

pllim: Anyway, support for mask is added (for now).

pllim:
I think we can ignore the  RTD failure. Doc build is successful locally. I think RTD is trying to use outdated cache and there is no way I can wipe a PR build.

eteq:
LGTM now too!  So I'll go ahead and merge.  I see the remaining questions were deferred to elsewhere.

pllim:
Thanks for the reviews, everyone!

PR #252, opened 2020-02-26
[ENH] Analytic integrals

### LINKED ISSUE #123: Implement analytic integration for 1D models
Unlikely #88 would happen anytime soon, so just go ahead an implement it here for the models that we use. No point to use trapezoid integration for simple things like `Box1D`.

pllim:
Fix #123 

**TODO**

- [x] Finish implementation.
- [x] Polish change log.
- [x] Implement astropy/astropy#7593 -- See astropy/astropy#10015
- [x] Add and fix tests.

### DIFF
--- a/synphot/models.py
+++ b/synphot/models.py
@@ -2,6 +2,7 @@
 """Spectrum models not in `astropy.modeling`."""
 
 # STDLIB
+import math
 import warnings
 from copy import deepcopy
 from functools import partial
@@ -134,6 +135,9 @@ class BlackBody1D(Fittable1DModel):
 
         return bbflux.value
 
+    def integrate(self, *args):
+        return const.sigma_sb * (self.temperature ** 4)
+
 
 class BlackBodyNorm1D(BlackBody1D):
     """Create a normalized :ref:`blackbody spectrum <synphot-planck-law>`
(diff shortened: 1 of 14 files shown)


synphot/models.py
```diff
@@ -587,6 +618,16 @@ def sampleset(self, factor_step=0.1, **kwargs):
 
         return np.asarray(w)
 
+    # TODO: Remove unit hardcoding when we use model with units natively.
+    def integrate(self, x):
+        # TODO: Fix the formula, apply units, check for n_models>1
```
pllim: Something is not quite right with this one. @eslavich , since you suggested the first derivative of Gaussian for this, maybe you can check my rusty calculus? Or @mcara ?

mcara: Corrected equation:
```python
return self.amplitude * (a2 - a1)
```

pllim: Thank you!

synphot/models.py
```diff
@@ -611,6 +618,16 @@ def sampleset(self, factor_step=0.1, **kwargs):
...
+        sig2 = self.sigma * self.sigma
+        dx_min = min(x) - self.x_0
+        dx_max = max(x) - self.x_0
+        a1 = dx_min * np.exp(-0.5 * dx_min * dx_min / sig2)
+        a2 = dx_max * np.exp(-0.5 * dx_max * dx_max / sig2)
+        return -1 * self.amplitude * (a2 - a1) / sig2
```
mcara: To make things easier...
```suggestion
        return self.amplitude * (a2 - a1)
```

pllim:
I might need to detour into astropy/astropy#7593 for a bit so I can convert units between integrated flux easily...

pllim:
Thank you, @stscirij , @mcara , @rendinam, and @eslavich for your help with the `RickerWavelet1D` integral!

PR #264, opened 2020-05-19
DOC: Clarify vegaspec in convert_flux

### LINKED ISSUE #262: Documentation request for Vegamag normalization
The current version of the documentation on the [normalization function](https://synphot.readthedocs.io/en/latest/api/synphot.spectrum.BaseSourceSpectrum.html#synphot.spectrum.BaseSourceSpectrum.normalize) does not make it clear that a Vega SourceSpectrum object must be supplied with the `vegamag=` keyword to normalize by vegamags. 

The existence of the SourceSpectrum.from_vega() method (and docstrings in synphot.units.convert_flux) implied to me that the keyword was optional and the system would run SourceSpectrum.from_vega() if a vega spectrum was not supplied (and maybe it should? It would greatly simplify users' code)

pllim:
Fix #262 

Is this clearer, @ariedel ?

### DIFF
--- a/synphot/units.py
+++ b/synphot/units.py
@@ -187,8 +187,9 @@ def convert_flux(wavelengths, fluxes, out_flux_unit, **kwargs):
         OBMAG and count, otherwise it is not needed.
 
     vegaspec : `~synphot.spectrum.SourceSpectrum`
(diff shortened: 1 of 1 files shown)


pllim:
Thanks for the reviews!

PR #268, opened 2020-05-23
Update build dependencies

migueldvb:
Currently the wheel and ~~numpy~~ packages are not required to build
synphot. Probably this is a remnant from the astropy-helpers era.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = ["setuptools>=30.3.0",
             "setuptools_scm",
-            "wheel",
             "oldest-supported-numpy",
             "extension-helpers"]
 build-backend = "setuptools.build_meta"


pllim:
I vaguely remember removing it but then encountering something that needed it, but I don't remember the details now. Anyway, I'll look into the CI failure first and come back to this.

pllim:
Hello and thank you for your interest in contributing!

Unfortunately, I don't think I can accept this PR because `wheel` is needed after all. If you look at the CI log of this PR, you would notice:
```
Using legacy setup.py install for psutil, since package 'wheel' is not installed.
```

I also consulted with @astrofrog and @Cadair , who wrote the APE 17 migration guide, which was what the `pyproject.toml` is based on. They said `wheel` is needed to install dependencies with wheels. (Please correct me if I misunderstood.)

Thanks anyway! It is truly appreciated.

PR #280, opened 2020-08-04
Re-enable automatic publication to PyPI

rendinam:
Previously the publication action was not compatible with the PEP517 build that is required by this project.


pllim:
Thanks (again)!

PR #282, opened 2020-09-15
add a target for the CI badge

eteq:
Very minor edit - I noticed the readme CI badge doesn't have a target, which means if you click on it you get just the badge.  I *think* this is now the right target.


pllim (approved):
Thanks! No need to run CI, so I canceled them.

p.s. Tsk tsk, you didn't use a fork.

PR #296, opened 2021-01-06
EXP: Undo hacks in tox and see what happens

### LINKED ISSUE #295: TST: Investigate tox behavior with C-extension
Follow up of #293 . Need to see why C-extension needs to be manually built before running tests in `tox`. Perhaps it is expected or perhaps it is a symptom of a larger problem. Someone with expertise in `tox` and C-extension packaging needs to have a look.

https://github.com/spacetelescope/synphot_refactor/blob/60be4e9a82ee81f8527cae2dcf516c6fc4274f23/tox.ini#L15-L18

https://github.com/spacetelescope/synphot_refactor/blob/60be4e9a82ee81f8527cae2dcf516c6fc4274f23/tox.ini#L55-L59

pllim:
### Description

This pull request is to investigate tox behavior.

Fixes #295

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-  pull_request_target:
+  pull_request:
   schedule:
     # Weekly Monday 6AM build
     # * is a special character in YAML so you have to quote this string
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_workflows.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-  pull_request_target:
+  pull_request:
```
pllim, line 5: Change this back before merge.

PR #310, opened 2021-04-23
Fix tox with C-extension

mcara:
For background info, see https://github.com/spacetelescope/synphot_refactor/issues/295 and related issues/PRs.

This PR avoids building the package in its own directory when running `tox`. This is done by changing import mode in `pytest` (see `setup.cfg`) and switching to absolute imports in test modules (which is a good thing anyway).

The only test that is failing is doc test when trying to import `synphot` due to relative import of `_astropy_init` - you will see this in failing tests.

CC: @pllim

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,8 +4,9 @@ testpaths = "synphot" "docs"
 norecursedirs = build docs/_build synphot/src
 astropy_header = true
 doctest_plus = enabled
+doctest_rst = true
 text_file_format = rst
-addopts = --doctest-rst --doctest-ignore-import-errors
+addopts = --import-mode append
 xfail_strict = true
 filterwarnings =
     ignore:can't resolve package from __spec__
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,7 @@ passenv = HOME WINDIR CC CI
 
 # Run the tests in a temporary directory to make sure that we don't import
 # package from the source tree
-changedir = {homedir}
+changedir = .tmp/{envname}
 
 # tox environments are constructued with so-called 'factors' (or terms)
 # separated by hyphens, e.g. test-devdeps-cov. Lines below starting with factor:
--- a/tox.ini
+++ b/tox.ini
@@ -12,10 +12,9 @@ isolated_build = true
 # Pass through the following environemnt variables which are needed for the CI
 passenv = HOME WINDIR CC CI
 
-# FIXME: Cannot do this, need to force C-ext to build below.
 # Run the tests in a temporary directory to make sure that we don't import
 # package from the source tree
-#changedir = .tmp/{envname}
+changedir = {homedir}
 
 # tox environments are constructued with so-called 'factors' (or terms)
 # separated by hyphens, e.g. test-devdeps-cov. Lines below starting with factor:
@@ -53,13 +52,12 @@ extras =
     alldeps: all
 
 commands =
-    # FIXME: Not sure why need this for tox to see the C-ext
-    python setup.py build_ext --inplace
-    # End of FIXME
     pip freeze
-    !cov: pytest --pyargs synphot {toxinidir}/docs {posargs}
-    cov: pytest --pyargs synphot {toxinidir}/docs --cov synphot --cov-config={toxinidir}/setup.cfg {posargs}
-    cov: coverage xml -o {toxinidir}/coverage.xml
+    # FIXME: Seems like using -c option with multiline Pytest is broken: see
+    # https://github.com/tox-dev/tox-travis/issues/146
+    # at the very least on CI - it runs fine locally on MacOSX.
+    !cov: pytest -c {toxinidir}/setup.cfg --pyargs synphot --rootdir={toxinidir} {posargs}
+    cov: pytest -c {toxinidir}/setup.cfg --pyargs synphot --rootdir={toxinidir} {posargs} --cov={envsitepackagesdir}/synphot --cov-config={toxinidir}/setup.cfg --cov-report xml:{toxinidir}/coverage.xml
 
 [testenv:codestyle]
 skip_install = true
--- a/synphot/tests/test_binning.py
+++ b/synphot/tests/test_binning.py
@@ -13,8 +13,8 @@ from astropy import units as u
 from astropy.utils.data import get_pkg_data_filename
 
 # LOCAL
-from .. import binning, exceptions, specio
-from ..utils import merge_wavelengths, generate_wavelengths
+from synphot import binning, exceptions, specio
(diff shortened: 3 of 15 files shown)


pllim:
With or without `import-mode` gave different errors. 😿

tox.ini
```diff
@@ -53,9 +52,6 @@ extras =
     alldeps: all
 
 commands =
-    # FIXME: Not sure why need this for tox to see the C-ext
-    python setup.py build_ext --inplace
-    # End of FIXME
     pip freeze
     !cov: pytest --pyargs synphot {toxinidir}/docs {posargs}
```
pllim, line 56: @jdavies-st , I do have `--pyargs` here. Am I using it wrong?

pllim:
I converted this to draft PR for now. FYI.

tox.ini
jdavies-st: With `changedir` commented out above, because you are running `pytest --pyargs` from within a directory that has `synphot` as a subdir, every `import synphot` will import it from the local checkout, not the installed location, following normal python import rules that always imports the local dir thing before checking python path.

setup.cfg
```diff
@@ -4,8 +4,9 @@ testpaths = "synphot" "docs"
 norecursedirs = build docs/_build synphot/src
 astropy_header = true
 doctest_plus = enabled
+doctest_rst = true
```
pllim, line 7: Huh, why this addition, is `addopts = --doctest-rst` not enough?

tox.ini
```diff
@@ -12,10 +12,9 @@ isolated_build = true
 # Pass through the following environemnt variables which are needed for the CI
 passenv = HOME WINDIR CC CI
 
-# FIXME: Cannot do this, need to force C-ext to build below.
 # Run the tests in a temporary directory to make sure that we don't import
 # package from the source tree
-#changedir = .tmp/{envname}
+changedir = {homedir}
```
pllim, line 17: Does it have to be `homedir`? Did `changedir = .tmp/{envname}` not work?

tox.ini
```diff
@@ -36,7 +35,7 @@ deps =
 
     # The oldestdeps factor is intended to be used to install the oldest versions of all
     # dependencies that have a minimum version.
-    oldestdeps: numpy==1.16.*
+    oldestdeps: numpy==1.17.*
```
pllim, line 38: Please don't bump the minversion here. This is out of scope.

tox.ini
mcara: Does it matter?

tox.ini
```diff
@@ -53,13 +52,12 @@ extras =
...
-    python setup.py build_ext --inplace
-    # End of FIXME
     pip freeze
     !cov: pytest --pyargs synphot {toxinidir}/docs {posargs}
-    cov: pytest --pyargs synphot {toxinidir}/docs --cov synphot --cov-config={toxinidir}/setup.cfg {posargs}
-    cov: coverage xml -o {toxinidir}/coverage.xml
+    # FIXME: Seems like using -c option with multiline Pytest is broken: see
+    # https://github.com/tox-dev/tox-travis/issues/146
+    # at the very least on CI - it runs fine locally on MacOSX.
+    cov: pytest --rootdir={toxinidir} -c {toxinidir}/setup.cfg --pyargs synphot --cov={envsitepackagesdir}/synphot --cov-config={toxinidir}/setup.cfg {posargs} --cov-report xml:{toxinidir}/coverage.xml
```
pllim, line 60: Can you explain why the changes here?

tox.ini
pllim: Yes, `homedir` might be pre-populated with stuff. `.tmp/{envname}` is clean.

mcara: The point is that tests do run in the install directory unlike what it was before - the source directory. Whether they run in
```
home/.tox/py39.../lib/python3.9/site-packages/synphot
```
or 
```
chdir .tmp/py39env
../../home/.tox/py39env/lib/python3.9/site-packages/synphot
```
does it make a difference?

pllim: I'd rather you change it to `.tmp/{envname}` anyway, if either one works. That is what `astropy` is using and easier for me to copy paste from upstream in the future.

tox.ini
mcara: Not really. It seems like the directory in which `pytest` is going to run depends on a combination of factors such as `rootdir`, location of the config for pytest, etc. I was just lucky enough to guess the right incantation.

tox.ini
mcara: Done

tox.ini
mcara: Done

mcara:
@pllim, in https://github.com/spacetelescope/synphot_refactor/blob/c6623f452e72f3cd636fa4a3b8b4bedfe51f8956/setup.cfg#L8 why do you use this option `--doctest-ignore-import-errors`?

pllim:
It was from astropy/pytest-doctestplus#108 but I don't remember now if it was added as a precaution or I needed it. You can try remove it and see.

pllim (commented):
I feel like this might have added some stuff that is unnecessary. I'll play with this a bit and report back. Thanks for investigating!

pllim (changes requested):
So... I look at the logs here and noticed that it is not actually running the doctest in the `docs` folder.

And... Purely by accident, I found out that the presence of `conftest.py` in the root directory messes up `tox`. So, simply by removing it (#313), `tox` picks up the installed version again *and* is also running the doctest. Though I hate to lose the really informative `pytest` header, I think removing `conftest.py` at the root dir is the more correct way to go. What do you think, @mcara and @jdavies-st ?

Also see spacetelescope/jdaviz#583 .

mcara:
> Not sure about the other stuff you added.

Like what? `-c` ? I do not think I added anything else.

pllim:
Well, look at the diff.

mcara:
What specifically about diff worries you? I just removed your workaround that you have marked with `# FIXME: Not sure why need this for tox to see the C-ext` and the code below that was building the package in place and as the result the test was run in the source directory. I thought you wanted to be able to run tests in the install directory. If that is not what you want - feel free to close this PR.

mcara:
That is, this PR allows running unit tests on installed package instead of the source directory with the package build in place (as it is currently done). All tests seem to be passing and running from the installed directory.

However, I see that #314 is achieving the same with other added benefits like picking up `rst` tests in the source directory. I think that one has better results and for this reason I am closing this PR in favor of #314

pllim:
@mcara , this PR really helped shined the light on why things weren't working. So even though this wasn't merged, thank you very much for your hard work on this! I'll be sure to give you co-author credit for #314.

PR #311, opened 2021-04-23
MNT: Remove _astropy_init.py, replace deprecated np.float

pllim:
MNT: Remove _astropy_init.py

MNT: Replace deprecated np.float in doc.

### Description

This pull request is a pre-requisite for #310


mcara (commented):
6 lines out. That's great!

PR #312, opened 2021-05-04
Use absolute imports in unit tests

mcara:
This PR contains only those changes from https://github.com/spacetelescope/synphot_refactor/pull/310 that pertain to switching unit tests from relative to absolute imports.

The changes to `tox.ini` will be addressed in https://github.com/spacetelescope/synphot_refactor/pull/310


pllim (approved):
Thank you so much!

PR #313, opened 2021-05-07
EXP: Is root conftest.py the culprit?

pllim:
### Description

This pull request is to address ...

Fixes #<Issue Number>


pllim:
tl;dr -- yes, it is, but #314 also fixes the problem without having to remove this file.

PR #331, opened 2022-05-11
Separate VEGAMAG and OBMAG

### LINKED ISSUE #327: Redefine VEGAMAG and OBMAG so astropy can tell them apart
https://github.com/spacetelescope/synphot_refactor/blob/e51d7605349eebea0921c7d72da674b0af7d6db9/synphot/units.py#L52-L55

As stated in https://github.com/astropy/astropy/pull/13158#discussion_r858962549 , the new `astropy.modeling.models.Schechter1D` takes magnitude as parameter but currently it cannot tell VEGAMAG and OBMAG apart. @mhvk advised the following:

<hr/>

[The] problem is that they now are both just aliases to `u.mag`, so the units system cannot know what to do. I think best might be to treat it a bit like ST and AB, and define units `VEGA` and `OB` - which both being units without a known number, and then define `VEGAMAG = u.mag(VEGA)` and `OBMAB = u.mag(OB)`. Then one can no longer just mix and match those magnitude systems:

```python
>>> from astropy import units as u
>>> VEGA = u.def_unit('VEGA')
>>> OB = u.def_unit('OB')
>>> VEGAMAG = u.mag(VEGA)
>>> OBMAG = u.mag(OB)
>>> (5 * VEGAMAG - 2.5 * VEGAMAG).to(u.one)
<Quantity 0.1>
>>> (5 * VEGAMAG - 2.5 * OBMAG).to(u.one)
UnitConversionError: ("'VEGA / OB' and '' (dimensionless) are not convertible",
'Did you perhaps subtract magnitudes so the unit got lost?')
```

Note that of course the user can still shoot themselves in the foot by forgetting what band a magnitude is in...

<hr/>

To go with this patch, relevant documentation and tests might also need updating. And this might be a breaking change, so should not be in a bugfix release.

pllim:
### Description

This pull request is to fix #327 

*This is breaking change!* Unit string will change from, say, `'VEGAMAG'` to `'mag(VEGA)'`. Also the behavior also changes because it is now a magnitude object, not just "plain Quantity."

### TODO

- [x] Make tests pass.
- [x] Test `stsynphot` against this PR branch. See spacetelescope/stsynphot_refactor#162
- [x] Need blessing from ETC. cc @vglaidler @ariedel @cdsontag 
  - Contacted them on 2022-05-17
  - Response from Adric: *I tried it on some of our regression tests that use vegamags and abmags (we have none that use OBmags). Our multimission/normalization and jwst/workbooks test suites all passed without complaint. Pandeia is compatible with your change... I think the PR is OK to merge. It passes all the relevant tests.* (2022-05-23)

### Blocked by (SOLVED)

<details>

@mhvk , I am trying to follow your advice at https://github.com/astropy/astropy/pull/13158#discussion_r858962549 but I am unable to get the new magnitude definitions to work with existing unit conversion equivalencies. What am I missing?

https://github.com/spacetelescope/synphot_refactor/blob/bd229608b8a2977ca1fad9cf2a2d5abb4bfcae80/synphot/units.py#L244-L245

https://github.com/spacetelescope/synphot_refactor/blob/bd229608b8a2977ca1fad9cf2a2d5abb4bfcae80/synphot/units.py#L291-L293

https://github.com/spacetelescope/synphot_refactor/blob/bd229608b8a2977ca1fad9cf2a2d5abb4bfcae80/synphot/units.py#L128

```
.../synphot/units.py in _convert_flux(wavelengths, fluxes, out_flux_unit, area, vegaspec)
    289             area = area * AREA
    290
--> 291         out_flux = fluxes.to(
    292             out_flux_unit,
    293             equivalencies=spectral_density_count(wavelengths, area))

.../astropy/units/quantity.py in to(self, unit, equivalencies, copy)
    844             # Avoid using to_value to ensure that we make a copy. We also
    845             # don't want to slow down this method (esp. the scalar case).
--> 846             value = self._to_value(unit, equivalencies)
    847         else:
    848             # to_value only copies if necessary

.../astropy/units/quantity.py in _to_value(self, unit, equivalencies)
    798         if not self.dtype.names or isinstance(self.unit, StructuredUnit):
    799             # Standard path, let unit to do work.
--> 800             return self.unit.to(unit, self.view(np.ndarray),
    801                                 equivalencies=equivalencies)
    802

.../astropy/units/function/core.py in to(self, other, value, equivalencies)
    255             try:
    256                 # when other is not a function unit
--> 257                 return self.physical_unit.to(other, self.to_physical(value),
    258                                              equivalencies)
    259             except UnitConversionError as e:

.../astropy/units/core.py in to(self, other, value, equivalencies)
   1128             return UNITY
   1129         else:
-> 1130             return self._get_converter(Unit(other),
   1131                                        equivalencies=equivalencies)(value)
   1132

.../astropy/units/core.py in _get_converter(self, other, equivalencies)
   1044         # if that doesn't work, maybe we can do it with equivalencies?
   1045         try:
-> 1046             return self._apply_equivalencies(
   1047                 self, other, self._normalize_equivalencies(equivalencies))
   1048         except UnitsError as exc:

.../astropy/units/core.py in _apply_equivalencies(self, unit, other, equivalencies)
   1002                     pass
   1003                 try:
-> 1004                     scale1 = tunit._to(unit)
   1005                     scale2 = funit._to(other)
   1006                     return make_converter(scale1, b, scale2)

AttributeError: 'MagUnit' object has no attribute '_to'
```

</details>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 - Dropped support for Python 3.6 and 3.7. Minimum supported Python
   version is now 3.8. [#330]
 
+- OBMAG and VEGAMAG are no longer interchangeable. [#331]
+
 1.1.1 (2021-11-18)
 ==================
 
--- a/docs/synphot/spectrum.rst
+++ b/docs/synphot/spectrum.rst
@@ -48,7 +48,7 @@ for flux conversion)::
     <Quantity [6.62607015e-24, 6.62607015e-23] FNU>
     >>> area = 45238.93416 * units.AREA  # HST
     >>> sp(wave, flux_unit=units.OBMAG, area=area)  # doctest: +FLOAT_CMP
-    <Quantity [-21.52438718,-21.52438718] OBMAG>
+    <Magnitude [-21.52438718,-21.52438718] mag(OB)>
 
 .. _synphot_reddening:
 
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -458,7 +458,7 @@ class Observation(BaseSourceSpectrum):
         if flux_unit == u.count or flux_unit_name == units.OBMAG.to_string():
             val = self.countrate(area, binned=False, wavelengths=wavelengths)
 
-            if flux_unit.decompose() == u.mag:
+            if flux_unit == units.OBMAG:
                 eff_stim = (-2.5 * np.log10(val.value)) * flux_unit
             else:
                 eff_stim = val
--- a/synphot/tests/test_units.py
+++ b/synphot/tests/test_units.py
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Test units.py module.
 
-.. note:: VEGAMAG conversion is tested in test_spectrum.py.
+.. note:: VEGAMAG conversion is tested in test_spectrum_source.py.
 
 .. note:: spectral_density_integrated is tested in astropy>=4.1.
 
@@ -13,6 +13,7 @@ import pytest
 
 # ASTROPY
 from astropy import units as u
+from astropy.tests.helper import assert_quantity_allclose
 
 # LOCAL
 from synphot import exceptions, units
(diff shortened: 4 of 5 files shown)


mhvk:
@pllim - my memory of our previous discussion is too vague. Is it possible to give a minimal example that now fails but should work?

pllim:
> now fails but should work

I don't know how to provide a minimal example. I have a couple of equivalencies that have been working for years. Now they crash with the traceback above when I apply your suggested changes to the magnitude units. I thought magnitude unit as you suggested would be a drop-in replacement but it appears not so.

mhvk:
OK, I think I can see why `spectral_density_count` equivalency would not work for `mag(OB)`. What's needed is a conversion to `OB`, and let the `Magnitude` take care of the logs. So, something like:
```
PHOTLAM = u.def_unit(
    'photlam', u.photon / (u.cm**2 * u.s * u.AA),
    format={'generic': 'PHOTLAM', 'console': 'PHOTLAM'})
_ob = u.def_unit('OB')
OBMAG = u.mag(_ob)
factor = 100  # calculate as in equivalency
equiv = [(PHOTLAM, OBMAG.physical_unit, lambda x: x*factor, lambda x: x/factor)]
(10*OBMAG).to(PHOTLAM, equivalencies=equiv)
# <Quantity 1.e-06 PHOTLAM>
```

From a quick look, it seems that in your `spectral_density_count` equivalency, the conversion routines now become identical for counts and OBMAG.

pllim:
Thanks, @mhvk ! That is very helpful. 😸

PR #336, opened 2022-07-24
Enforce little endian when reading Numpy test data file

olebole:
Numpy data (written with `ndarray.tofile()` are usually system dependent; at least they depend on the byte order. This makes the `test_countrate_neg_leak()` [fail on big endian platforms (IBM zSystems/s390x)](https://buildd.debian.org/status/fetch.php?pkg=synphot&arch=s390x&ver=1.1.1-2&stamp=1658536121&raw=0). This patch enforces little endian on all platforms with `np.fromfile()` so the test passes there as well. With this patch, [the package passes on s390x](https://buildd.debian.org/status/fetch.php?pkg=synphot&arch=s390x&ver=1.1.1-3&stamp=1658641250&raw=0).

### DIFF
--- a/synphot/tests/test_observation.py
+++ b/synphot/tests/test_observation.py
@@ -454,7 +454,7 @@ def test_countrate_neg_leak():
         package='synphot.tests'))
     binset = np.fromfile(get_pkg_data_filename(
         os.path.join('data', 'stis_fuv_f25ndq2_binset.bin'),
-        package='synphot.tests'))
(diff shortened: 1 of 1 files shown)


pllim (approved):
Thanks! LGTM but I'll let the CI run. Even though this look good, I might not come back to this in a few days.

Do you think it is beneficial to add a s390x job like astropy here or is that overkill?

pllim:
p.s. Also let me know if you need a new release. Thanks!

PR #338, opened 2022-09-20
Remove AstroConda installation instructions

ojustino:
### Description

I noticed that the docs ask for Python >=3.8 while AstroConda is only compatible with Python <=3.7. As such, the installation option that uses the AstroConda channel doesn't work.

If you like this change, I can propose the same one for `stsynphot`.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -53,11 +53,7 @@ packages:
 
 You can install **synphot** using one of the following ways:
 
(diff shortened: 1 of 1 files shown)


pllim (approved):
Hmm. I still pushed to astroconda recipes in the last release though I think I am going to stop going forward, so might as well. Thanks!

pllim:
No need for CI.

PR #342, opened 2022-11-30
Modifying Box1D model to allow step input 

### LINKED ISSUE #340: Operations with Box1D spectral element significantly slower than with GaussianFlux1D
### Description

Not entirely sure if this is a real issue or I'm misunderstanding some synphot internals, but I have noticed that Observation operations when using a bandpass based on Box1D are systematically slower than equivalent operations with a bandpass based on a Gaussian element. 

### Expected behavior

Both sets of operations to take approximately the same amounts of time.

### Steps to Reproduce

MWE example below. Specifying integration_type has no effect in either case. 

```python
from synphot import Observation, SourceSpectrum, SpectralElement, units
from synphot.models import BlackBodyNorm1D, GaussianFlux1D, Box1D
import numpy as np
s = SourceSpectrum(BlackBodyNorm1D, temperature=5000)
l = s.avgwave()
BW = l*0.1
FWHM = BW*np.sqrt(np.log(2)/np.pi)*2
box = SpectralElement(Box1D, x_0=l, width=BW)
gauss = SpectralElement(GaussianFlux1D, mean=l, fwhm=FWHM)

obs_box = Observation(s, box)
obs_gauss = Observation(s, gauss)

flux_box = obs_box.integrate()
flux_gauss = obs_gauss.integrate()
print(flux_box - flux_gauss) # -0.0006330283047517593 ph / (cm2 s)
```

Timing tests with ipython on two different systems:
```
%timeit obs_box = Observation(s, box)
#sys 1: 458 ms ± 43.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
#sys 2: 106 ms ± 368 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit obs_gauss = Observation(s, gauss)
#sys 1: 62.8 ms ± 2.57 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
#sys 2: 5.77 ms ± 16.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit flux_box = obs_box.integrate()
#sys 1: 265 ms ± 22.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
#sys 2: 27.1 ms ± 250 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit flux_gauss = obs_gauss.integrate()
#sys 1: 48 ms ± 5.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
#sys 2: 4.33 ms ± 42 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### System Details

sys1: 
macOS-11.7-x86_64-i386-64bit
Python 3.10.8 (main, Oct 12 2022, 02:49:24) [Clang 13.0.0 (clang-1300.0.29.3)]
Numpy 1.23.4
astropy 5.1
Scipy 1.8.1
Matplotlib 3.5.3
Synphot 1.1.1

sys2:
Linux-4.15.0-176-generic-x86_64-with-glibc2.27
Python 3.10.2 (main, Feb 22 2022, 10:06:51) [GCC 7.5.0]
Numpy 1.22.2
astropy 5.0.1
Scipy 1.8.0
Matplotlib 3.5.1
Synphot 1.1.1

dsavransky:
### Description

This pull request is to address #340. Adds step input to Box1D model while preserving prior behavior as default and adds covering unit test. 

Fixes #340

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@
 
 - OBMAG and VEGAMAG are no longer interchangeable. [#331]
 
+- ``Box1D`` model now takes optional ``step`` input to allow user control
+  over the generate sampleset. Default behavior maintains backwards 
+  compatibility. [#340]
+
 1.1.1 (2021-11-18)
 ==================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@
 
 - ``Box1D`` model now takes optional ``step`` input to allow user control
   over the generated sampleset. Default behavior maintains backwards 
-  compatibility. [#340]
+  compatibility. [#342]
 
 1.1.1 (2021-11-18)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,7 @@
 - OBMAG and VEGAMAG are no longer interchangeable. [#331]
 
 - ``Box1D`` model now takes optional ``step`` input to allow user control
-  over the generate sampleset. Default behavior maintains backwards 
+  over the generated sampleset. Default behavior maintains backwards 
   compatibility. [#340]
 
 1.1.1 (2021-11-18)
--- a/synphot/models.py
+++ b/synphot/models.py
@@ -192,10 +192,14 @@ class Box1D(_models.Box1D):
         (nominally Angstrom). Defaults to 0.01
 
     """
-    def __init__(self, step=0.01, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
+
+        if "step" in kwargs:
+            self.step = kwargs.pop("step")
+        else:
+            self.step = 0.01
 
         super(Box1D, self).__init__(*args, **kwargs)
-        self.step = float(step)
 
     @staticmethod
     def _calc_sampleset(w1, w2, step, minimal):
--- a/synphot/models.py
+++ b/synphot/models.py
@@ -185,7 +185,18 @@ class Box1D(_models.Box1D):
     """Same as `astropy.modeling.functional_models.Box1D`, except with
     ``sampleset`` defined.
 
+    Parameters
+    ----------
+    step : float
+        Distance of first and last points w.r.t. bounding box. In default units
+        (nominally Angstrom). Defaults to 0.01
+
     """
+    def __init__(self, step=0.01, *args, **kwargs):
+
+        super(Box1D, self).__init__(*args, **kwargs)
+        self.step = float(step)
+
     @staticmethod
     def _calc_sampleset(w1, w2, step, minimal):
         """Calculate sampleset for each model."""
@@ -196,19 +207,23 @@ class Box1D(_models.Box1D):
 
         return arr
 
-    def sampleset(self, step=0.01, minimal=False):
+    def sampleset(self, step=None, minimal=False):
         """Return ``x`` array that samples the feature.
 
         Parameters
         ----------
-        step : float
-            Distance of first and last points w.r.t. bounding box.
+        step : float, optional
+            Distance of first and last points w.r.t. bounding box. If None,
+            set to attribute ``step``.
 
         minimal : bool
             Only return the minimal points needed to define the box;
             i.e., box edges and a point outside on each side.
 
         """
+        if step is None:
+            step = self.step
+
         if ASTROPY_LT_5_0:
             w1, w2 = self.bounding_box
         else:
--- a/synphot/tests/test_spectrum_misc.py
+++ b/synphot/tests/test_spectrum_misc.py
@@ -152,20 +152,18 @@ class TestWaveset:
         np.testing.assert_array_equal(bp2.waveset.value, w_true)
         np.testing.assert_array_equal(bp3.waveset.value, w_true)
 
-    def test_box1d_coarse_step(self):
-        bp = SpectralElement(Box1D, x_0=2000, width=1, step=0.1)
-        w = bp.waveset.value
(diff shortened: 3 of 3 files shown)


synphot/models.py
```diff
@@ -185,7 +185,18 @@ class Box1D(_models.Box1D):
...
+    ----------
+    step : float
+        Distance of first and last points w.r.t. bounding box. In default units
+        (nominally Angstrom). Defaults to 0.01
+
     """
+    def __init__(self, step=0.01, *args, **kwargs):
+
+        super(Box1D, self).__init__(*args, **kwargs)
+        self.step = float(step)
```
pllim, line 198: Is an explicit float cast really needed here?

dsavransky: Possibly not, but what could it hurt?  I'm happy to remove if you wish.

synphot/tests/test_spectrum_misc.py
```diff
@@ -152,6 +152,21 @@ def test_box1d(self):
...
+
+        np.testing.assert_array_equal(w, w_true)
+        np.testing.assert_allclose(
+            w[([0, 1, -2, -1], )], bp.model.sampleset(minimal=True))
+
+        # Make sure scale does not change waveset
+        bp2 = bp * 2
+        bp3 = 0.5 * bp
+        np.testing.assert_array_equal(bp2.waveset.value, w_true)
+        np.testing.assert_array_equal(bp3.waveset.value, w_true)
```
pllim, line 168: Looks like this was copied from `test_box1d` and can be removed.

synphot/models.py
```diff
@@ -185,7 +185,18 @@ class Box1D(_models.Box1D):
...
     ``sampleset`` defined.
 
+    Parameters
+    ----------
+    step : float
+        Distance of first and last points w.r.t. bounding box. In default units
+        (nominally Angstrom). Defaults to 0.01
+
     """
+    def __init__(self, step=0.01, *args, **kwargs):
```
pllim, line 195: This will break API. I wonder if it is safer to extract `step` from `kwargs`.

pllim: Might also have to register it here

https://github.com/spacetelescope/synphot_refactor/blob/a5fa6533bdcb58bb43d8ee43526d15c9817292bf/synphot/spectrum.py#L74

```python
'Box1D': {'amplitude': 'flux', 'x_0': 'wave', 'width': 'wave', 'step': 'wave'},
```

pllim:
Please also add a change log entry in https://github.com/spacetelescope/synphot_refactor/blob/master/CHANGES.rst . Thanks!

pllim:
Weird that CI didn't kick off. Let me close/reopen and see.

synphot/models.py
dsavransky: Sorry - could you please point me to what exactly in the API will break? keywords should automatically be popped from **kwargs at runtime.  There should be no distinction between ``Box1D(**{'amplitude':1, 'x_0':1, 'width':1, 'step':1})`` and ``Box1D(amplitude=1, x_0 = 1, width=1, step=1)`` and any combination of separate keywords/kw dict.

pllim: Since `Box1D` signature does accept `*args`, I am worried the first positional argument might be passed in as `step` with this change.

synphot/models.py
pllim: Let's remove it. Unless you want something like `Box1D(..., step='0.1')` to magically work.

CHANGES.rst
```diff
@@ -12,6 +12,10 @@
 
 - OBMAG and VEGAMAG are no longer interchangeable. [#331]
 
+- ``Box1D`` model now takes optional ``step`` input to allow user control
+  over the generate sampleset. Default behavior maintains backwards
```
pllim, line 16:
```suggestion
  over the generated sampleset. Default behavior maintains backwards 
```

CHANGES.rst
```diff
@@ -12,6 +12,10 @@
 
 - OBMAG and VEGAMAG are no longer interchangeable. [#331]
 
+- ``Box1D`` model now takes optional ``step`` input to allow user control
+  over the generate sampleset. Default behavior maintains backwards 
+  compatibility. [#340]
```
pllim, line 17:
```suggestion
  compatibility. [#342]
```

pllim (commented):
Change log suggestions below. Thanks! 😸

synphot/models.py
dsavransky: Ah. You're absolutely right.  I'll re-implement as you suggest. Thanks.

pllim:
Oh no, I think we got hit by tox 4.0 too.

pllim:
Hello. Can you please rebase to pick up https://github.com/spacetelescope/synphot_refactor/pull/343 ? Thanks!

pllim:
Thank you! LGTM now but I need to know that stakeholders like @vglaidler, @ariedel, and co. would be okay with this. In the meantime, I ask for your patience. 😸

cdsontag (commented):
Hi @pllim !  I have no concerns since this is backwards compatible, but our two likeliest reviewers are @ariedel and @isaacspitzer who cannot look at this until they get back next week.

ariedel (approved):
This change looks good. The ETC does not use Box1D and I have verified it does not break as a result of this change. If we were going to use Box1D, I think we'd run into the same initial bug and greatly benefit from being able to control the step size.

pllim:
Thank you, all!

PR #353, opened 2023-04-05
refactor toxenvs and use OpenAstronomy workflows

zacharyburnett:
### Description

This pull request consolidates CI jobs in GitHub Actions to use the OpenAstronomy reusable workflow, and also updates the Tox configuration to use new keyword aliases. Additionally, adds the `-xdist` factor to run `pytest-xdist`

image

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -16,7 +16,7 @@ permissions:
   contents: read
 
 jobs:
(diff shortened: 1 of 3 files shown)


pllim:
Seems like overkill. This isn't a high traffic package and the current workflow works.

pllim:
There is conflict now. And I don't have time to review this, so let's just close.

PR #370, opened 2024-01-05
MNT: upgrade upload/download-artifact GitHub actions

neutrinoceros:
### Description

Hi, version 4 of upload-artifact and download-artifact were recently released.
Staying on old version is not viable long term because GitHub will sooner or later drop support for the old versions of node that these depend on.
Since migration to v4 is not as straightfoward as bumping the version number, I'm helping all astropy-related packages byt manually performing the necessary changes.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -55,14 +55,9 @@ jobs:
         path: ./wheelhouse/*.whl
 
   build_dist:
-    name: Source dist and publish
+    name: Source dist
     runs-on: ubuntu-latest
-    needs: build_wheels
-    environment:
-      name: release
-      url: https://pypi.org/p/synphot
-    permissions:
-      id-token: write  # IMPORTANT: mandatory for trusted publishing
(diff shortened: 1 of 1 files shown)


pllim:
Hello! It looks like I did not configure the publish workflow to ever run for a PR. Maybe let me fix that and then you can rebase, so we know it works before merge? Thanks for your patience!

neutrinoceros:
sure, by all means !

neutrinoceros:
@pllim thanks for updating this branch. I think the failure is unrelated, can you re-trigger the job ?

pllim:
@neutrinoceros , they didn't fail. I cancelled them because I realized I needed to enable artifact upload/download also for non-release events. I am still debating with myself whether I should disable them again after it is proven or not.

pllim:
Ah... I mixed the release env stuff into dist build job too. I need to ponder this.

pllim (approved):
Thanks! I added a commit to clean up my own tech debt. I think this is GTG now.

pllim:
Also, yes, I thought about migrating to OpenAstronomy template but resisted because I got the Trusted Publishing already set up here and that does not play well with templating (yet).

PR #373, opened 2024-02-08
update actions version

zacharyburnett:
### Description

This pull request is to address ...

Fixes #<Issue Number>

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check base branch
-      uses: actions/github-script@v6
+      uses: actions/github-script@v7
       if: github.event_name == 'pull_request'
       with:
         script: |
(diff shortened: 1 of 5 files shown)


pllim:
What would be more future-proof is to instead submit a PR with `dependabot.yml` and let the bot do all the upgrades. Are you willing to do that? Thanks!

.github/workflows/ci_workflows.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check base branch
-      uses: actions/github-script@v6
+      uses: actions/github-script@v7
```
pllim, line 24: I don't remember if github-script from v6 to v7 requires also other updates to go with it. You might have to check the change log.

zacharyburnett:
> What would be more future-proof is to instead submit a PR with `dependabot.yml` and let the bot do all the upgrades. Are you willing to do that? Thanks!

Yes I think that's a great idea. I'll write up a PR

pllim:
Thanks! I will close this without merge then so we can test to see if the bot works when it exists.

pllim:
xref #377

PR #384, opened 2024-03-15
BUG/API: read_fits_spec now more flexible but at a cost

### LINKED ISSUE #372: specio.read_spec assumes wavelength and flux are always in column 1 and column 2
### Description

If the wavelength column in a spectrum is not column 1 and the flux/throughput column is not column 2, the units are read incorrectly.

### Expected behavior

Loading a file should grab the wavelength unit of the wavelength column, and the flux unit of the flux column.

### Actual behavior

The ETC has a file describing the spectral traces for NIRISS SOSS, wherein the columns are x (pixels), wavelength (microns), and trace (pixels). When reading it in, the code crashed because specio.read_spec produced a file where the wavelength was supposedly in pixels, and the trace was supposedly in microns.

### Steps to Reproduce

1. grab this file: https://github.com/spacetelescope/pandeia_data/blob/master/jwst/niriss/wavepix/jwst_niriss_soss-96-ord1_trace_20160927172406.fits
2. Observe the header units:
```python
import astropy.io.fits as fits

with fits.open("jwst_niriss_soss-96-ord1_trace_20160927172406.fits") as hdulist:
    print(hdulist[1].header)
```
3. Read in with synphot:
```python
import astropy.units as u
import synphot

microns = u.def_unit('MICRONS', u.micron)
pixels = u.def_unit("PIXELS", u.pixel)

u.add_enabled_units([microns, pixels])

spectrum = synphot.spectrum.SpectralElement.from_file("jwst_niriss_soss-96-ord1_trace_20160927172406.fits", flux_col="trace")
```
4. Observe a crash: `UnitConversionError: 'Unit("PIXELS")' is not a scaled version of 'Unit("Angstrom")'`. Note that the wavelength unit shouldn't be pixels; it should be microns.

I have tracked the issue down to specio.read_fits_spec [line 185](https://github.com/spacetelescope/synphot_refactor/blob/d07fe2ae73b87904021e02471cc2b449ee5c1ad0/synphot/specio.py#L185) where in we read the data by field name, and the units from the header always as TUNIT1 and TUNIT2.
In this particular case, the wavelength column was 2, and the "flux" column was 3, so TUNIT2 and TUNIT3 should have been read.
I have experimentally written some code to fix that, which I can PR.

### System Details

macOS-10.16-x86_64-i386-64bit
Python 3.11.4 (main, Jul  5 2023, 09:00:44) [Clang 14.0.6 ]
Numpy 1.26.3
astropy 6.0.0
Scipy 1.11.4
Matplotlib 3.7.2
Synphot 1.2.1

pllim:
### Description

This pull request is to address the problem of `read_fits_spec` unnecessarily hardcoding `TUNIT1` and `TUNIT2`. However, in making it more flexible, some breaking changes are unavoidable. See change log added here.

Fixes #372 

### After merge

- [ ] Check to see if a compatibility PR is needed for `stsynphot` downstream.

### DIFF
--- a/synphot/tests/test_specio.py
+++ b/synphot/tests/test_specio.py
@@ -158,18 +158,24 @@ def test_read_nonstandard_fits_cols_01(tmp_path):
     trace = np.array([0, 0.5, 1, 0.9, 0])
     coldefs = fits.ColDefs([
         fits.Column(name="X", format="I", array=pix),
-        fits.Column(name="WAVELENGTH", format="E",
+        fits.Column(name="Wavelength", format="E",
                     unit=wav.unit.to_string(format="fits"), array=wav.value),
-        fits.Column(name="TRACE", format="E", array=trace)])
+        fits.Column(name="Trace", format="E", array=trace)])
     hdulist = fits.HDUList([
         fits.PrimaryHDU(),
         fits.BinTableHDU.from_columns(coldefs)])
     outfile = str(tmp_path / "jwst_niriss_soss_trace.fits")
     hdulist.writeto(outfile, overwrite=True)
 
-    tr = SpectralElement.from_file(outfile, flux_col="TRACE")
-    assert_quantity_allclose(tr.waveset, wav)
-    assert_quantity_allclose(tr(wav), trace, atol=1e-7)
+    # Make sure column names are still case insensitive.
+    for (wave_col, flux_col) in (
+            ("Wavelength", "Trace"),
+            ("WAVELENGTH", "TRACE"),
+            ("wavelength", "trace")):
+        tr = SpectralElement.from_file(
+            outfile, wave_col=wave_col, flux_col=flux_col)
+        assert_quantity_allclose(tr.waveset, wav)
+        assert_quantity_allclose(tr(wav), trace, atol=1e-7)
 
 
 def test_read_nonstandard_fits_cols_02(tmp_path):
--- a/synphot/tests/test_specio.py
+++ b/synphot/tests/test_specio.py
@@ -3,8 +3,6 @@
 
 # STDLIB
 import os
-import shutil
-import tempfile
 
 # THIRD-PARTY
 import numpy as np
@@ -15,10 +13,12 @@ from astropy import units as u
 from astropy.io import fits
 from astropy.tests.helper import assert_quantity_allclose
 from astropy.utils.data import get_pkg_data_filename
-from astropy.utils.exceptions import AstropyUserWarning
+from astropy.utils.exceptions import (
+    AstropyUserWarning, AstropyDeprecationWarning)
 
 # LOCAL
 from synphot import exceptions, specio, units
+from synphot.spectrum import SpectralElement
 
 
 @pytest.mark.remote_data
@@ -55,7 +55,6 @@ class TestReadWriteFITS:
     """Test read/write FITS spectrum."""
     def setup_class(self):
         self.epsilon = 0.00031
-        self.outdir = tempfile.mkdtemp()
         self.wave = np.array([1000.0, 2000.0, 2000.0 + self.epsilon, 3000.0,
                               4000.0, 5000.0], dtype=np.float64) * u.AA
         self.flux = np.array([0.1, 100.2, 10.0, 0.0, 6.5, 1.2],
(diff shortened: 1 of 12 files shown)


synphot/tests/test_specio.py
```diff
@@ -149,5 +150,55 @@ def test_exceptions(self):
             specio.write_fits_spec(
                 outfile, self.wave, np.arange(6), overwrite=True)
 
-    def teardown_class(self):
-        shutil.rmtree(self.outdir)
+
+def test_read_nonstandard_fits_cols_01(tmp_path):
```
pllim, line 154: This is the example given by @ariedel .

synphot/tests/test_specio.py
```diff
@@ -149,5 +150,55 @@ def test_exceptions(self):
...
+        fits.BinTableHDU.from_columns(coldefs)])
+    outfile = str(tmp_path / "jwst_niriss_soss_trace.fits")
+    hdulist.writeto(outfile, overwrite=True)
+
+    tr = SpectralElement.from_file(outfile, flux_col="TRACE")
+    assert_quantity_allclose(tr.waveset, wav)
+    assert_quantity_allclose(tr(wav), trace, atol=1e-7)
+
+
+def test_read_nonstandard_fits_cols_02(tmp_path):
```
pllim, line 175: This is the example given by @talister .

synphot/units.py
```diff
@@ -295,6 +295,18 @@ def validate_unit(input_unit):
             output_unit = THROUGHPUT
         elif input_unit_lowcase == 'jy':
             output_unit = u.Jy
+        elif input_unit_lowcase == "flam":
```
pllim, line 298: If there are anymore weird units that STScI uses in any table column `synphot` needs to parse, we have to hardcode the exception rule here, or fix the data.

Now that we go though `astropy.table.QTable`, it parses all the columns. AFAIK, there is no way to skip. This is the price we pay for a more flexible table format.

ariedel:
I have tried this branch, and I am concerned about the removal of the case insensitivity. The ETC web interface uses synphot to read in user-uploaded spectra; I worry that this would lead to rejection of a large number of spectra and a worse user experience.
(This branch has also revealed that the ETC's own data files are very heterogeneous when it comes to capitalization (@rizeladiaz).)

Part of the reason the original issue appeared is that the ETC is now trying to expand our use of Synphot to reading in files that aren't actually spectra or spectral elements (quantum yield curves, dispersion files) in a unit-aware way. If you have an hour or half hour of time to chat with us about the best ways to accomplish this, it would be much appreciated.

pllim:
> case insensitivity

This is a side effect of using Unified I/O in `astropy`. All the parsing is done now in https://github.com/astropy/astropy/blob/main/astropy/io/fits/connect.py . I was surprised too that it is case-sensitive, so maybe this is a bug that should be fixed upstream. But I will have to ask my `astropy` colleagues.

> reading in files that aren't actually spectra or spectral elements (quantum yield curves, dispersion files) in a unit-aware way

If you have no plans to use these objects within `synphot` for further analysis, you might be abusing the parser here and better off using something more generic like `astropy`.

pllim:
p.s. I opened https://github.com/astropy/astropy/issues/16221 but now that I have thought about it as I typed it up, maybe I can force the case-insensitivity here. Will report back.

pllim:
@ariedel , I managed to make it case insensitive again. Is this good enough for you?

ariedel:
This is much better. It's potentially revealed more problems in STScI's TRDS dataset, but I don't think Synphot should be responsible for them @rizeladiaz 
* I can't load the Bruzual77 spectra from $PYSYN_CDBS/grid/bz77; all of them yield `ValueError: Function units cannot be written in fits format. Only generic, unscaled, latex, latex_inline, unicode, console are supported.` I have not taken the time to track down what that error means yet or how it could be worked around.
* I also can't normalize anything by the 2MASS H bandpass in $PYSYN_CDBS/comp/nonhst/2mass_h_001_syn.fits because it has a unit of "THROUGHPUT" which isn't recognized by Astropy or Synphot. (I'll need to specially define it as equivalent to u.dimensionless_unscaled in our normalization code.)

pllim:
Ah, yes, I didn't test this in stsynphot. Maybe I should.

ariedel:
I'm pretty sure it's because the flux unit in those bz77 files is "STMAG". 
The ETC is trying to use `syn.spectrum.SourceSpectrum.from_file(filepath)` to read it, rather than any explicit STSynphot function; the error is coming from line 201: `subhdu.header[key] = newval.to_string("fits")`.

pllim:
If it is just a matter of weird unit strings, we can add exception rules here if they are not too weird.

ariedel (approved):
I have now tested this, and approve.

pllim:
FWIW I restarted the devdeps job at https://github.com/spacetelescope/stsynphot_refactor/actions/runs/8338326996 and it still passed. But if you find anything amiss, please open new issue(s). Thanks for the review!

PR #387, opened 2024-04-05
MNT: numpy 2 and astropy 6.1 compat, bump minversions

### LINKED ISSUE #386: MNT: Drop Python 3.9
To be consistent with astropy v6.1

When it is done here, also do the same for `stsynphot`.

pllim:
### Description

This pull request is to address impending numpy 2.0 release.

Fixes #386

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -66,13 +66,13 @@ jobs:
 
           - name: Test without optional deps
             os: ubuntu-latest
-            python: 3.9
-            toxenv: py39-test
+            python: '3.10'
+            toxenv: py310-test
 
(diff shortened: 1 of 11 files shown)


pllim:
@ariedel , does ETC team want to take this for a spin before I merge? This builds synphot C-ext against numpy 2.x ABI but should still be compatible with numpy 1.x to some degree.

pllim:
Please let me know if you also need an accompanying stsynphot PR though I don't see why because stsynphot does not have C-extension.

ariedel (approved):
The ETC still uses numpy 1.x, but I've taken this branch for a spin and the ETC functions perfectly well with the changes.

PR #388, opened 2024-05-31
[SCSB-155] build with Numpy 2.0 release candidate

zacharyburnett:
### Description

resolves [SCSB-155](https://jira.stsci.edu/browse/SCSB-155)

Numpy 2.0 is scheduled for release on June 19th. Due to the C ABI updates, Python projects that build C extensions are required to build against Numpy 2.0 in order to use Numpy 2.0 in runtime.
This is backwards-compatible with Numpy versions in runtime as far back as `1.26`

this PR pins `numpy>=2.0.0rc2` in `build-system.requires`
> [!NOTE]
> this PR was generated automatically by ``batchpr`` :robot:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
 requires = ["setuptools>=30.3.0",
             "setuptools_scm",
-            "numpy>=2.0.0rc1"]
+            "numpy>=2.0.0rc2"]
 build-backend = "setuptools.build_meta"


pllim (commented):
But why? What is the difference between RC1 and RC2? Why not just wait for numpy 2 stable that is coming out in a few weeks?

zacharyburnett:
> But why? What is the difference between RC1 and RC2? Why not just wait for numpy 2 stable that is coming out in a few weeks?

well, I would like to have a build with *some* version of the 2.0 ABI before the full Numpy 2.0 release, to head off installation issues for users who don't have Numpy pinned, and so we can test along our dependency web; unfortunately, it looks like even if you build against numpy 2.0, if you depend on an upstream project built against numpy 1.x there will be issues

pllim:
> if you depend on an upstream project built against numpy 1.x there will be issues

Yes, that is true whether it is RC1 or RC2. The whole chain needs to be either all on 1.x ABI or all on 2.x ABI.

zacharyburnett:
good point, in this case since it's already `rc1` it doesn't really matter. Sorry, I was confusing this with another project that was using `oldest-supported-numpy`; too many PRs! 😅

pllim:
No worries!

pllim:
Yes, oldest-supported-numpy is incompatible with numpy 2.0 stuff.

PR #403, opened 2025-01-06
Force flux_unit to be unit instance

pllim:
Force flux_unit to be unit instance instead of a str to be compatible with astropy v7.1

xref https://github.com/astropy/astropy/pull/17586

### DIFF
--- a/synphot/spectrum.py
+++ b/synphot/spectrum.py
@@ -484,7 +484,7 @@ class BaseSpectrum:
                 raise exceptions.SynphotError(
                     'flux_unit cannot be used with unitless spectrum')
             else:
-                self._validate_flux_unit(flux_unit, wav_only=True)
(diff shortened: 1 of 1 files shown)


mhvk:
A bit less obviously also an improvement in the code than for `jdaviz` - while you are at it, does it make sense to also re-assign in the one other case where `self._validate_flux_unit` is called? (I just did a code search, so am not sure whether it is in fact needed)

pllim:
You mean here?

https://github.com/spacetelescope/synphot_refactor/blob/63a63f24240b6549668b052bfa911520134c78d7/synphot/spectrum.py#L1144-L1146

In this case, `pval` is already a Quantity, so we do not have any problem with string.

mhvk:
Yes, sorry I didn't look better before commenting!

PR #435, opened 2026-03-03
Use default Vega spectrum for flux conversions

### LINKED ISSUE #433: Why does `normalize` not load the Vega spectrum itself if needed?
### Description
In some cases, in particular in the `normalize` method of a `SourceSpectrum` the code needs a Vega spectrum, which that user needs to pass in. The doctoring of the `normalize` methods points to the docstring of `convert_flux` which point to the doctoring of `from_vega`, which then allows me to read the Vega spectrum from the database.

There is no decision by the user involved in any way. Why doesn't the code simply get a Vega spectrum automatically, if it is needed?

### Additional context
For performance, it might even be useful to load the spectrum only once and cache it or keep it in the module level variable.

In the very unlikely case that the user want to use a different spectrum for Vega, they could still do that be changing the relevant configuration value or (if cached in a module attribute) simply fill in that attribute themselves.

hamogu:
There is only one Vega in the sky and most users will simply use the spectrum given the the database for their conversions. Thus, there is no reasons to already require them to explicitly put in a spectrum. With this PR, they still can, but if not, the conversion functions will try to load the default Vega spectrum if it's available.

Fixes #433

### DIFF
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -24,8 +24,7 @@ from synphot.models import Empirical1D
 from synphot.spectrum import (
     BaseSourceSpectrum,
     SourceSpectrum,
-    SpectralElement,
-    _get_cached_vega,
+    SpectralElement
 )
 
 __all__ = ['Observation']
@@ -475,7 +474,9 @@ class Observation(BaseSourceSpectrum):
         # This is basically effstim(self)/effstim(Vega)
         if flux_unit_name == units.VEGAMAG.to_string():
             if vegaspec is None:
-                vegaspec = _get_cached_vega()
+                from synphot import spectrum
+                spectrum.lazy_load_vega()
+                vegaspec = spectrum.Vega
 
             num = self.integrate(wavelengths=wavelengths)
             den = (vegaspec * self.bandpass).integrate(
@@ -649,10 +650,7 @@ class Observation(BaseSourceSpectrum):
             Flux is converted to this unit for plotting.
             If not given, internal unit is used.
 
-        area : float or `~astropy.units.quantity.Quantity`
-            See :func:`~synphot.units.convert_flux`.
-
-        vegaspec : `~synphot.spectrum.SourceSpectrum`
+        area, vegaspec
             See :func:`~synphot.units.convert_flux`.
 
         kwargs : dict
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -472,7 +472,7 @@ class Observation(BaseSourceSpectrum):
         if flux_unit_name == units.VEGAMAG.to_string():
             if not isinstance(vegaspec, SourceSpectrum):
                 from synphot import spectrum
-                spectrum.lazy_load_vega()
+                spectrum._lazy_load_vega_with_exception()
                 vegaspec = spectrum.Vega
 
             num = self.integrate(wavelengths=wavelengths)
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -21,8 +21,12 @@ from astropy.utils.exceptions import (AstropyUserWarning,
 # LOCAL
 from synphot import binning, exceptions, units, utils
 from synphot.models import Empirical1D
-from synphot.spectrum import (BaseSourceSpectrum, SourceSpectrum,
-                              SpectralElement)
+from synphot.spectrum import (
+    BaseSourceSpectrum,
+    SourceSpectrum,
+    SpectralElement,
+    _get_cached_vega,
+)
 
 __all__ = ['Observation']
 
@@ -471,10 +475,8 @@ class Observation(BaseSourceSpectrum):
         # This is basically effstim(self)/effstim(Vega)
         if flux_unit_name == units.VEGAMAG.to_string():
             if vegaspec is None:
-                vegaspec = SourceSpectrum.from_vega()
-            # Vega data file missing, so from_vega above returns None.
-            if not isinstance(vegaspec, SourceSpectrum):
-                raise exceptions.SynphotError('Vega spectrum is missing.')
+                vegaspec = _get_cached_vega()
+
             num = self.integrate(wavelengths=wavelengths)
             den = (vegaspec * self.bandpass).integrate(
                 integration_type='trapezoid')
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -21,11 +21,8 @@ from astropy.utils.exceptions import (AstropyUserWarning,
 # LOCAL
 from synphot import binning, exceptions, units, utils
 from synphot.models import Empirical1D
-from synphot.spectrum import (
-    BaseSourceSpectrum,
-    SourceSpectrum,
-    SpectralElement
-)
+from synphot.spectrum import (BaseSourceSpectrum, SourceSpectrum,
+                              SpectralElement)
 
 __all__ = ['Observation']
 
@@ -473,7 +470,7 @@ class Observation(BaseSourceSpectrum):
         # Special handling of VEGAMAG.
         # This is basically effstim(self)/effstim(Vega)
         if flux_unit_name == units.VEGAMAG.to_string():
-            if vegaspec is None:
+            if not isinstance(vegaspec, SourceSpectrum):
                 from synphot import spectrum
                 spectrum.lazy_load_vega()
                 vegaspec = spectrum.Vega
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -470,6 +470,9 @@ class Observation(BaseSourceSpectrum):
         # Special handling of VEGAMAG.
         # This is basically effstim(self)/effstim(Vega)
         if flux_unit_name == units.VEGAMAG.to_string():
+            if vegaspec is None:
+                vegaspec = SourceSpectrum.from_vega()
+            # Vega data file missing, so from_vega() returns None.
             if not isinstance(vegaspec, SourceSpectrum):
                 raise exceptions.SynphotError('Vega spectrum is missing.')
             num = self.integrate(wavelengths=wavelengths)
@@ -644,8 +647,12 @@ class Observation(BaseSourceSpectrum):
             Flux is converted to this unit for plotting.
             If not given, internal unit is used.
 
-        area, vegaspec
+        area:
             See :func:`~synphot.units.convert_flux`.
+        vegaspec : `~synphot.spectrum.SourceSpectrum`
+            If this kwarg is present, the given Vega spectrum to use for conversion
+            of VEGAMAG. If not present, it is automatically loaded from
+            :func:`~synphot.spectrum.SourceSpectrum.from_vega`.
 
         kwargs : dict
             See :func:`synphot.spectrum.BaseSpectrum.plot`.
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -647,7 +647,7 @@ class Observation(BaseSourceSpectrum):
             Flux is converted to this unit for plotting.
             If not given, internal unit is used.
 
-        area : float
+        area : float or `~astropy.units.quantity.Quantity`
             See :func:`~synphot.units.convert_flux`.
 
         vegaspec : `~synphot.spectrum.SourceSpectrum`
--- a/synphot/observation.py
+++ b/synphot/observation.py
@@ -472,7 +472,7 @@ class Observation(BaseSourceSpectrum):
         if flux_unit_name == units.VEGAMAG.to_string():
             if vegaspec is None:
                 vegaspec = SourceSpectrum.from_vega()
-            # Vega data file missing, so from_vega() returns None.
+            # Vega data file missing, so from_vega above returns None.
             if not isinstance(vegaspec, SourceSpectrum):
                 raise exceptions.SynphotError('Vega spectrum is missing.')
             num = self.integrate(wavelengths=wavelengths)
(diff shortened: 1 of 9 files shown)


hamogu (commented):
I would like to add a test to ensure that the warnings are still shown when the data is NOT available.
I can cook something up (e.g. change the conf object to point to a non-existent path before the test and then change it back after), but before I make it too complicated, I thought I'll just ask is there is a fixture or something already that I overlooked to test without loading data?

**edited to add thought for how to implement that**

synphot/observation.py
```diff
@@ -470,6 +470,9 @@ def effstim(self, flux_unit=None, wavelengths=None, area=None,
         # Special handling of VEGAMAG.
         # This is basically effstim(self)/effstim(Vega)
         if flux_unit_name == units.VEGAMAG.to_string():
+            if vegaspec is None:
+                vegaspec = SourceSpectrum.from_vega()
```
pllim: Should we do some sort of caching internally? 

Maybe akin to https://github.com/spacetelescope/stsynphot_refactor/blob/2431029d13c4286e26af9b33ec71bae9a288b1e6/stsynphot/spectrum.py#L33 ?

I mean, would you find the way Vega is handled in stsynphot useful?

pllim:
Thanks for the PR! I have pinged STScI ETC team for opinions. Once we agree on the high-level concept, we can then fine-tune the implementation details.

synphot/tests/test_observation.py
```diff
@@ -337,12 +337,14 @@ def test_effstim_vegamag(self):
             self.obs.effstim(flux_unit=units.VEGAMAG, vegaspec=vspec).value,
             ans, rtol=0.001)
 
-    @pytest.mark.parametrize('flux_unit', [u.mag, units.VEGAMAG])
```
pllim, line 340: I have to think about this. `u.mag` should still be tested.

synphot/observation.py
```diff
@@ -470,6 +470,9 @@ def effstim(self, flux_unit=None, wavelengths=None, area=None,
...
+            # Vega data file missing, so from_vega above returns None.
```
pllim: Are you sure it returns None if data file is missing? I think it will just crash based on

https://github.com/spacetelescope/synphot_refactor/blob/39f01bb5e0eee44875bb3ec951c26a0de63f365b/synphot/spectrum.py#L1383-L1384

ariedel:
I like this change. It makes it easier to use synphot generically, and there's still a way to override the choice of Vega spectrum if you want to (the ETC does)

synphot/observation.py
hamogu: It's in fact crashing earlier with a "FileNotFound" error. I think I happened to test it from an environment that also has `stsynphot` which overwrites the locations where it's looking for something.

synphot/observation.py
hamogu: I'm suggesting a slightly different mechanism. stsynphot is probably used with a local database in most cases, but synphot defaults to getting the Vega spectrum through https, so I want to load it lazily.

synphot/tests/test_observation.py
hamogu: It is tested elsewhere, but it doesn't cause an exception any longer, thus I saw little point in testing it in a test called "test_XXX_exceptions".
However, some parts of this test are now resurrected to test for the exception message when a database is not available.

pllim:
Remote data job is failing. Do you see the same failure when you run the test suite with `--remote-data` flag locally?

Update: I can reproduce this locally. I can investigate.

synphot/tests/test_observation.py
```diff
@@ -337,12 +338,37 @@ def test_effstim_vegamag(self):
...
+        ],
+    )
+    def test_effstim_exceptions(self, flux_unit, message):
+        """By default, the Vega spectrum is lazily loaded when needed. To test the error
+        when the Vega spectrum cannot be loaded, we temporarily
+        modify the package configuration to point to a non-existent Vega file.
+        """
+        conf_vega_file = conf.vega_file
+        try:
+            conf.vega_file = ""
```
pllim: It is a bit cleaner to use https://docs.astropy.org/en/latest/api/astropy.time.Conf.html#astropy.time.Conf.set_temp here. 😄

synphot/tests/test_spectrum_source.py
```diff
@@ -74,15 +66,40 @@ def test_flux_conversion_vega(in_q, out_u, ans):
...
 
+@pytest.mark.parametrize(
+    ("in_q", "out_u", "identical_ans"),
+    [
+        (_flux_photlam, units.VEGAMAG, False),
+        (_flux_vegamag, units.PHOTLAM, False),
+        (_flux_jy, units.FLAM, True),
+    ],
+)
+def test_flux_conversion_vega_different_spec(in_q, out_u, identical_ans):
```
pllim, line 84: I would change the `test_flux_conversion_vega` test name to maybe like `test_flux_conversion_vega_builtin`. Otherwise if I ever want to do `-k test_flux_conversion_vega`, I will always pick up `test_flux_conversion_vega_different_spec` also whether I want to or not.

synphot/tests/test_spectrum_source.py
```diff
@@ -504,15 +520,24 @@ def test_exceptions(self):
...
         # Zero flux
         sp = SourceSpectrum(Const1D, amplitude=0)
         with pytest.raises(exceptions.SynphotError):
             sp.normalize(100 * u.ct, band=self.abox, area=_area)
 
+    def test_exception_missing_vegaspec(self):
+        """This is the same logic as "test_exception" but pulled out
+        into a separate test because it needs more setup."""
+        conf_vega_file = conf.vega_file
+        try:
```
pllim: Same comment as https://github.com/spacetelescope/synphot_refactor/pull/435/changes#r2885178525

synphot/spectrum.py
```diff
@@ -26,8 +26,42 @@
...
+]
+
+
+Vega = (
+    None  # This is defined at the end of this module to avoid circular import issues.
+)
+"""Default Vega spectrum. Used for conversions to and from VEGAMAG when no Vega spectrum is given.
+
+This is lazily loaded when needed using :func:`SourceSpectrum.from_vega()`.
+There are several mechanisms to use a different spectrum for Vega:
```
pllim, line 46:
```suggestion
There are several mechanisms to use a different spectrum for Vega:

```

synphot/spectrum.py
```diff
@@ -26,8 +26,42 @@
...
+- Set ``synphot.conf.vega_file``
+- Set ``synphot.spectrum.Vega`` before it is used
+- Provide Vega spectrum as argument to conversion functions that require it (e.g., ``units.convert_flux``).
```
pllim:
```suggestion
- Provide Vega spectrum as argument to conversion functions that require it (e.g., :func:`synphot.units.convert_flux`).
```

synphot/spectrum.py
```diff
@@ -26,8 +26,42 @@
...
+    if Vega is None:
+        try:
+            Vega = SourceSpectrum.from_vega()
+        except Exception as e:
+            new_message = (
+                f"Need Vega spectrum for conversion but cannot load: {e.args[0]}\n"
+            )
+            new_message += "Fix loading or provide Vega spectrum as argument."
+            e.args = (new_message,) + e.args[1:]
+            raise exceptions.SynphotError((new_message,) + e.args[1:])
```
pllim: * Do we need to combine this `new_message` twice?
* Should we `raise ... from None`?

synphot/spectrum.py
```diff
@@ -1069,10 +1103,7 @@ def normalize(self, renorm_val, band=None, wavelengths=None, force=False,
 
             # VEGAMAG
             if renorm_unit_name == units.VEGAMAG.to_string():
-                if not isinstance(vegaspec, SourceSpectrum):
-                    raise exceptions.SynphotError(
-                        'Vega spectrum is missing.')
-                stdspec = vegaspec
+                stdspec = _get_cached_vega()
```
pllim: We should respect user override here, right?
```suggestion
                if not isinstance(vegaspec, SourceSpectrum):
                    stdspec = _get_cached_vega()
                else:
                    stdspec = vegaspec            
```

synphot/spectrum.py
```diff
@@ -26,8 +26,42 @@
...
+    "BaseUnitlessSpectrum",
+    "SpectralElement",
+    "Vega",
+]
+
+
+Vega = (
+    None  # This is defined at the end of this module to avoid circular import issues.
+)
+"""Default Vega spectrum. Used for conversions to and from VEGAMAG when no Vega spectrum is given.
```
pllim, line 43: If you really want this docstring exposed,  you might have to do `.. autodata::` directive in the RST doc. I don't including it in `__all__` is enough because it is not a class nor function.

synphot/tests/test_observation.py
hamogu: I was looking for a test decorator or context manager like that, but didn't find one and decided that the small number of cases didn't warrant writing one. I didn't know that astropy itself already has that.
I'm learning something every time!

hamogu:
I'm happy to continue working on this, but maybe it's easier if you fix that? If I do, then I have to wait again for the workflow run to be approved, and then check again etc.

All tests pass locally (even with the `--remote-data` option), but I don't have tox and pre-commit set up for this repro, so the unused imports slipped through.

pllim:
No worries. I can clean up. Thanks!

synphot/observation.py
```diff
@@ -644,7 +649,10 @@ def plot(self, binned=True, wavelengths=None, flux_unit=None, area=None,
             Flux is converted to this unit for plotting.
             If not given, internal unit is used.
 
-        area, vegaspec
+        area : float or `~astropy.units.quantity.Quantity`
```
pllim: Looking at this again, I don't think this change is needed. Otherwise, we have to change all similar docstrings to stay consistent. I think user can still get the info they need if the follow the link to `convert_flux` API doc. I am reverting this change here. FYI.

synphot/spectrum.py
```diff
@@ -26,8 +26,42 @@
...
+This is lazily loaded when needed using :func:`SourceSpectrum.from_vega()`.
+There are several mechanisms to use a different spectrum for Vega:
+
+- Set ``synphot.conf.vega_file``
+- Set ``synphot.spectrum.Vega`` before it is used
+- Provide Vega spectrum as argument to conversion functions that require it (e.g., :func:`synphot.units.convert_flux`).
+"""
+
+
+def _get_cached_vega():
```
pllim: Nothing wrong with this but I want to copy over the rest of the bells and whistles from `stsynphot.spectrum.load_vega` so in the future, I can have `stsynphot` just call this function to reduce code duplication. I will do that in the next commit.

pllim:
@hamogu , I made some changes to your PR branch. Please see if my changes still align with your original vision. I noticed that you changed some checks to `if vegaspec is None`, but I changed it back to `isinstance` check. This is because while your check is more correct, it does slightly change the logic. I think we want to just lazy load under the same conditions where it used to throw SynphotError.

@ariedel and/or @cdsontag , do you want to test this PR branch against your ETC stuff before or after merge? Please advise.

Thanks, all!

cdsontag:
Thanks for thinking of us @pllim , but (to my knowledge) we are not using the "synphot_refactor" repo.  If/when this goes into the synphot code we _are_ using, we can worry about breakage then (and we'll have time to deal with it).  That said, if @ariedel wants to give it a spin ahead of time, I have no objections.

synphot/spectrum.py
```diff
@@ -26,8 +26,74 @@
...
+    if vegafile is None:
+        vegafile = conf.vega_file
+
+    with conf.set_temp("vega_file", vegafile):
+        try:
+            Vega = SourceSpectrum.from_vega(**kwargs)
+        except Exception as e:
+            Vega = None
+            new_err_message = (
+                f"Need Vega spectrum for conversion but cannot load: {e.args[0]}\n"
```
hamogu: This error message was good when this was all internal. But now "layy_load_vega" is a public facing function. If I use that to get a Vega spectrum (say, to just quickly get a Vega spectrum to show to an undergrad or to just get a random Spectrum object to test out something else), when "Need Vega Spectrum for conversion" is not the right text any more and "provide Vega spectrum as an argument" is plain wrong - this function doesn't take a spectrum as an argument, but a file name (or URL). 

I'm not quite sure how to solve this: The message could go back into the individual functions that need it (as it was before this PR), but pulling it out and implementing it only once but will all best and whistles (a useable default, can override, prints helpful and instructive error message) was kind of the point here.

Or, we could fall back to the old maxim that every problem in CS can be solved with one more level of indirection: Let this function just issue a generic exception and add a second, private function "_get_lazy_loaded_Vega_spectrum_for_VEGAMAG_conversion" that will call this function and wrap any exceptions with the more useful, instructive text.

Or, a bit simpler, make this function private (at least don't include it in `__all__`) and if users want a Vega, they can still do it the old way.

pllim: The reason I want to expose this is because I eventually want to use it downstream myself (https://github.com/spacetelescope/stsynphot_refactor/issues/222). It feels wrong for me to design a package to use another package's "hidden" function.

hamogu: I get the motivation, then then either the error message must be made more generic (and thus less helpful for people who simple want to convert something into VEGAMAG) or we need to wrap it in some way, just to modify the error message. I would be fine with either, it's really our call as the maintainer which one you prefer.

pllim: How about this?

```python
f"Failed to load a built-in Vega spectrum from {vegafile}: {e.args[0]}\n"
"Fix loading in lazy_load_vega() function or provide it directly via vegaspec keyword."
```

hamogu: Except that the `lazy_load_vega` doesn't have the "vegaspec" keyword, so that would be:
"If you are converting to/from VEGAMAG, you can also provide a Vega spectrum directly via the vegaspec keyword".

(Of course, 99% of the users will never use this function directly, but only indirectly when converting something, so maybe I'm too much of a stickler here.)

pllim: It is a good point. I pushed another commit out to address your concern. Does it satisfy your needs or it is too convoluted? Please advise. Thanks.

hamogu:
I think this is great now and should be merged.
(Since I don't have write permission I can't the GH feature to click "Approve".)

pllim:
Since ETC does not need to test this before merge, let's do it then. Thanks, all!

ariedel:
I did actually test this branch (relatively late version in development) and it did work with the ETC.

pllim:
Good to hear. Thanks for checking, @ariedel !

